PR #33, opened 2015-12-03
Adding support for Wind River Linux in support of Cisco devices

adamleff:
Wind River Linux is the underlying OS for Cisco Nexus and IOS-XR
devices, and is essentially a Red Hat variant.  These changes allow
train to connect to these devices and run commands properly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 train-*.gem
+Gemfile.lock
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -36,7 +36,7 @@ module Train::Extras
 
     OS = {
       'redhat' => %w{
-        redhat oracle centos fedora amazon scientific xenserver
+        redhat oracle centos fedora amazon scientific xenserver wrlinux
       },
       'debian' => %w{
         debian ubuntu linuxmint raspbian
--- a/lib/train/extras/os_detect_linux.rb
+++ b/lib/train/extras/os_detect_linux.rb
@@ -11,7 +11,7 @@
 require 'train/extras/linux_lsb'
 
 module Train::Extras
-  module DetectLinux
+  module DetectLinux # rubocop:disable Metrics/ModuleLength
     include Train::Extras::LinuxLSB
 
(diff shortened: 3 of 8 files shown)


chris-rock:
Hi Adam, thank you very much for this great addition. Do you think it may be possible to reuse the `redhatish_platform` & `redhatish_version` like https://github.com/chef/train/blob/adamleff/wrl-support/lib/train/extras/os_detect_linux.rb#L115?

adamleff:
Good afternoon, @chris-rock!

Wind River Linux is like RH in functionality, but in terms of how it describes itself name/version-wise, it's not that similar.  The os-release (and subsequent cisco-release file, if present) are in a multi-line but parseable format, and those files are customized by Cisco, as well.

So, `redhatish_platform` and `redhatish_version` doesn't make sense to use here.  This is why we used a similar technique in Ohai to detect the platform, platform_family, and platform_version for WRL rather than the "platformish" methods in Ohai as well:

https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L38-L45
https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L54-L64
https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb#L145-L159

arlimus:
Let's replicate the Ohai approach here; I originaly left out that code piece because we didn't test for Cisco yet. Love to get it back in!

adamleff:
@chris-rock / @arlimus - check out the latest commit I just pushed and tell me what you think, please! I started the unit test for os_detect_linux but did not do all of it.  I can certainly help write that coverage if needed but I figured a start is better than nothing, and I made sure to 100% cover the changes I was making.

Also test/unit/transports/local_test.rb was changed simply because minitest+mocha will complain for any object named "mock" so I had to change that to avoid a minitest failure.

chris-rock:
@adamleff looks very good from my side!

arlimus:
Really love the added tests, thank you so uch for all of it!!
A lot of improvements + Wind River Linux as a bonus ;)
Thank you @adamleff !
Cheers @chris-rock for the review :+1:

PR #36, opened 2015-12-04
Implemented WindowsFile#exist?

docwhat:
It was missing and not raising an error, which caused much confusion.

See https://github.com/chef/inspec/issues/288 for more info.


arlimus:
Fantastic clean fix for windows! :+1: 
Cheers @docwhat

PR #43, opened 2015-12-29
Support for local transport on Windows

chris-rock:
- ensures the windows detection is run via powershell
- prefix all commands with powershell if `local` transport is used
- adds a small script to run os detection with train only


arlimus:
Great addition for Windows, cheers @chris-rock !!

PR #44, opened 2015-12-30
introduce `mounted` as a seperate method to retrieve the content

chris-rock:


lib/train/extras/windows_file.rb
```diff
@@ -39,6 +39,10 @@ def link_path
       nil
     end
 
+    def mounted
+      nil
+    end
+
     def mounted?
       nil
     end
```
arlimus: how about just overwriting `mounted` and letting `mounted?` infer its status?

arlimus:
Looks great! Thank you @chris-rock :+1:

PR #47, opened 2016-01-13
Updating to the latest release of net-ssh to consume https://github.com/net-ssh/net-ssh/pull/280

tyler-ball:
\cc @chris-rock @arlimus


train.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |spec|
 
   spec.add_dependency 'json', '~> 1.8'
   spec.add_dependency 'mixlib-shellout', '~> 2.1'
-  spec.add_dependency 'net-ssh', '~> 2.9'
```
chris-rock: I propose we use `=> 2.9` since inspec works with ruby `1.9.3` as well

PR #50, opened 2016-01-15
bugfix: command wrapper always return nil

chris-rock:


arlimus:
:doh: :D cheers @chris-rock !!

PR #52, opened 2016-01-21
Copying https://github.com/test-kitchen/test-kitchen/pull/919 to this repo

tyler-ball:
\cc @chris-rock @arlimus 

I'm guessing we didn't see this here because something else in train was including timeout, but its still the right thing to do


chris-rock:
Thanks @tyler-ball for catching this!

PR #55, opened 2016-01-24
use negotiate auth for winrm and not basic_auth

mwrock:
winrm 1.6 provides support for NTLM/Negotiate auth so we should use that instead of `:plaintext` as the default transport. See http://www.hurryupandwait.io/blog/sane-authenticationencryption-arrives-to-ruby-based-cross-platform-winrm-remote-execution


chris-rock:
@mwrock Awesome. I cannot thank you enough for this feature! That makes live for our customers so much easier.

PR #57, opened 2016-01-25
Properly wrap commands in powershell for local backend

chris-rock:
- use the same approach as done in https://github.com/WinRb/WinRM/blob/master/lib/winrm/command_executor.rb
- add more testing on windows to see if `winrm` or `local` breaks

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,6 +89,25 @@ puts conn.file('/proc/version').content
 conn.close
 ```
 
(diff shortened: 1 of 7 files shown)


srenatus:
https://travis-ci.org/chef/train/jobs/104610133#L259 << The powershell unit test fails. Could you have a look?

PR #63, opened 2016-02-01
add `requiretty` workaround measures

srenatus:
This change includes a `chef-apply`able recipe stub that can be used to fix a user's `sudoers` config if the distribution defaults to `requiretty`.

```interactive
# TRAIN_SUDO_USER=chef chef-apply contrib/fixup-requiretty.rb
```

It tries to be cautious, since sudoer configurations are a delicate matter.


arlimus:
Fantastic workaround to a bugging can of problems!!
Thank you @srenatus :+1: :)

PR #70, opened 2016-02-14
complete rewrite of windows version detection

chris-rock:
The new Windows version detection is not based on Powershell anymore, it uses two core commands:
- `ver` is a command-line tool since MSDOS that outputs the version of the Operating System
- ~~`systeminfo` was introduced in Windows XP Professional and outputs more system details~~
- `wmic os get * /format:list` to retrieve the details of the operating system

Unfortunately we cannot rely on `ver` only, since build numbers and version are sometimes the same between Windows Client and Server. Therefore we cannot use this alone to distinguish between both operating systems.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -20,7 +20,8 @@ clone_folder: c:\projects\train
 clone_depth: 1
 
 cache:
-  - C:\Ruby21\bin\gem
+  - C:\Users\appveyor\.gem\ruby\2.1.0
+  - C:\Ruby21\lib\ruby\gems\2.1.0
 
 install:
   - systeminfo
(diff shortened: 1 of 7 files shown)


mwrock:
Might consider `run_wql` see https://github.com/WinRb/WinRM/blob/master/lib/winrm/command_executor.rb#L171

And querying `win32_operatingsystem`

mwrock:
Then you may just need one call

chris-rock:
@mwrock Great Idea. Is that working on CLI as well? We do not have WinRM available all the time. I try to find a solution that is very minimal and works remote and locally. It should have no specific dependency on PowerShell or any scripting.

mwrock:
Oh I thought this was all remote. Wmic would be the cli equivalent

chris-rock:
Thanks @mwrock for this great idea. `wmic` was a lifesaver and made the implementation much easier. Since the current implementation does not depend on Powershell anymore, it is much more stable and works awesome remote and locally.

arlimus:
Awesome improvement, and much simpler this way. Love the tests too @chris-rock :+1: 
Cheers @mwrock for your help :)

mwrock:
:hand:

PR #77, opened 2016-03-07
set default value for ssh compression to false

chris-rock:
Users have reported, that ssh compression does not work with all ssh servers. Lets default compression to false.


alexpop:
I see, we set compression to false for now and look into negotiating it between ssh client and server in the future?

alexpop:
:+1: 
Thanks Chris

chris-rock:
@alexpop that is the idea

PR #85, opened 2016-04-20
add -L to get stat for symlink

vjeffrey:
Fixes https://github.com/chef/inspec/issues/665

Tested on centos6, BSD.
Looks like aix is already using the lstat command, but I don't really know how to go about testing it. didn't change anything there.
When I tested, I confirmed that it now reads perms correctly for test referenced in the issue and also confirmed it did not modify the results of any other test.


srenatus:
Thank you, @vjeffrey 🎉

PR #92, opened 2016-04-28
Overhault file(...) and stat(...); point to destination of symlinks

arlimus:
As started by @alexpop and input from @vjeffrey , kudos!!

Right now, train's file implementation will always only look at the file under consideration, even if it is a symlink. If you want to check, if e.g. `/my/link` is being owned by `user`, it would only tell you if the symlink you specified is owned this way, but not if the destination it points to has the right permissions. It was also inconsistent internally, since `/dev/stdout` is `character_device?`, which cannot be determined from the source file you specified, but only from the destination it points to. Train did internally resolve this conflict for any type checks, but not for access permissions or anything else originating in `stat`.

After a lot of internal discussions we came to the conclusion, that the interface we are most interested in for testing any of these files would be pointing to the destination, not the source. If I wanted to know if `/dev/stdout` is a `symlink?` or owned by `root`, it would test that against wherever that file points to.

This also means, that there must be an accessor to give me the symlink only. While we have `link_path`, we also needed an object to run permissions checks. We decided against `link_source`, as it would only work consistently, when applied to links; similar to how `link_path` works right now:

```ruby
file('/dev/stdout').link_path => '/dev/fd/1'
file('/root').link_path => nil
```

Instead of `link_source` we opted for `source` with this mechanism:

```ruby
file('/dev/stdout').source => essentially file('/dev/stdout')
file('/root').link_path => essentially file('/root')
```

see:

train stat-v-lstat

There is still some work to be done on `link_path` to make it follow to the final target on all OS's (OS X is the issue right now, due to the lack of `readlink -f` by default).

### DIFF
--- a/lib/train/extras/file_common.rb
+++ b/lib/train/extras/file_common.rb
@@ -10,7 +10,7 @@ module Train::Extras
     # interface methods: these fields should be implemented by every
     # backend File
     %w{
-      exist? mode owner group link_target link_path content mtime size
+      exist? mode owner group link_path content mtime size
       selinux_label product_version file_version path
     }.each do |m|
       define_method m.to_sym do
@@ -126,8 +126,6 @@ module Train::Extras
     def target_type
       # Just return the type unless this is a symlink
       return type unless type == :symlink
-      # Get the link's target type, i.e. the real destination's type
-      return link_target.type unless link_target.nil?
       # Return unknown if we don't know where this is pointing to
       :unknown
     end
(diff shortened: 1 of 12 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @arlimus

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

alexpop:
Great improvement! Thanks Dom

PR #94, opened 2016-04-28
fix nil file paths

arlimus:
The problem with nil file paths is, that (a) they don't make any sense at all, but are contextually equal to `''` (empty string) and (b) they break a number of upstream methods like `File` in ruby. Since these are equivalent to `''` in all other regards (afaics), let's handle this case early on.

This is not trying to replace any other sources of input like numbers, non-string base types or even class instances. They should remain as e.g. custom file names could potentially prove dangerous. It's also easier to track down.

### DIFF
--- a/lib/train/extras/file_common.rb
+++ b/lib/train/extras/file_common.rb
@@ -20,7 +20,7 @@ module Train::Extras
 
     def initialize(backend, path, follow_symlink = true)
       @backend = backend
-      @path = path
+      @path = path || ''
       @follow_symlink = follow_symlink
     end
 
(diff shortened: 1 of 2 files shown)


chef-supermarket:
Hi. I am an automated pull request bot named Curry. There are commits in this pull request whose authors are not yet authorized to contribute to Chef Software, Inc. projects or are using a non-GitHub verified email address. To become authorized to contribute, you will need to sign the Contributor License Agreement (CLA) as an individual or on behalf of your company. [You can read more on Chef's blog.](https://www.getchef.com/blog/2014/06/23/changes-to-the-contributor-license-agreement-process)

## GitHub Users Who Are Not Authorized To Contribute

The following GitHub users do not appear to have signed a CLA:

* @arlimus

[Please sign the CLA here.](https://supermarket.chef.io/become-a-contributor)

PR #100, opened 2016-05-06
fix: winrm https listener is not configured anymore in appveyor

chris-rock:


alexpop:
Thanks for finding that out Chris!

PR #102, opened 2016-05-10
Update README.md

mcquin:
The ssh option for identity files is `:key_files`, not `:keys`.


arlimus:
Great catch, thank you @mcquin 👍

PR #107, opened 2016-05-14
Custom sudo command

jeremymv2:
This gives train the ability to execute a custom sudo command.
Customers whom utilize authorization software like Centrify will frequently employ an alternative custom command to 'sudo'. 

Default remains 'sudo' if `:sudo_command` option is not specified.


chris-rock:
Great addition @jeremymv2

PR #112, opened 2016-06-15
add missing keyboard-interactive authentication method

### LINKED ISSUE #111: Authentication with SSH Server on OSX is failing
https://github.com/chef/inspec/issues/788

chris-rock:
fixes #111

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -95,7 +95,7 @@ module Train::Transports
       end
 
       unless options[:password].nil?
-        options[:auth_methods].push('password')
+        options[:auth_methods].push('password', 'keyboard-interactive')
       end
 
       super


arlimus:
Kudos @chris-rock !!

The underlying discussion: This MR activates keyboard-interactive for SSH, which doesn't have much to do with its namesake. From the SSH config man page:

```
       KbdInteractiveDevices
              Specifies  the  list  of  methods to use in keyboard-interactive authentication.  Multiple method names must be comma-separated.  The default is to use the
              server specified list.  The methods available vary depending on what the server supports.  For an OpenSSH server, it may be zero or more  of:  ``bsdauth'',
              ``pam'', and ``skey''.
```

These mechanisms are required for some devices to authenticate. It also restores the default used in many SSH implementations.

The concern of the original implementation has been interactive i.e. blocking executions.

PR #114, opened 2016-06-16
adding support for vmware's esx server

Anirudh-Gupta:
added support for detecting esxi server. NO os_release or lsb_release file present. So only option is to go with uname. Doesn't come under linux or unix kernel. They have their own vmware proprietary kernel called VMKernel.


arlimus:
Thank you @Anirudh-Gupta !! Awesome to see ESX in here :+1: :smile:

PR #115, opened 2016-06-16
use train as gem name. Thanks @halo

chris-rock:
Special thanks to @halo who was so kind to hand over his gem name!


arlimus:
Huge kudos @halo !! :tada:

PR #116, opened 2016-06-17
ESX support

Anirudh-Gupta:
Added ESX file stat support by leveraging the already existing linux stat method.


chris-rock:
@Anirudh-Gupta thanks for the addition!

PR #120, opened 2016-06-27
do not log password in ssh connection output

marcparadise:
This addresses an immediate need for delivery CLIP, but a better longer-term approach may be to make something like: 

```
class ConnectionOptions < Hash 
  def to_s
    # find any  key => .*password.* with value type String  here and obscure them. 
  end
end
```
I didn't take that for this change because it'd be a mid-sized refactor. 

ping  @arlimus @srenatus

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -216,7 +216,9 @@ class Train::Transports::SSH
     #
     # @api private
     def to_s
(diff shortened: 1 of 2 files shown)


arlimus:
Awesome! Can you take care of the linting errors and we'll get this merged :)

srenatus:
LGTM! 🎉

srenatus:
Thanks a lot, @marcparadise!

PR #121, opened 2016-07-01
Allow requesting a PTY

### LINKED ISSUE #60: have net-ssh request a pty
The issue we face is basically described here: https://github.com/net-ssh/net-ssh/issues/103: when train tries to execute a command using `sudo`, and the target system's `sudoers` config contains `requiretty`, the call will fail with something like

```
$lib/train/extras/command_wrapper.rb:131:in `load': Sudo failed:  sudo: sorry, you must have a tty to run sudo (Train::UserError)
```

The way forward would be to request a pty, with [example code here](https://github.com/delano/rye/blob/master/lib/rye/box.rb#L886-L917).

@chris-rock @arlimus What do you think about that?

srenatus:
This change allows the user to request a pseudo-tty:

```
require 'train'
t = Train.create('ssh', host: '127.0.0.1', port: '2200', user: 'vagrant', password: 'vagrant', sudo: true, pty: true);
puts t.connection.run_command("ls /")
```

While this allows passing the `requiretty` sudoer defaults of certain
RedHat-ish distributions, it comes at the cost of having stderr and
stdout merged together. The change includes a warning for this.

The command above, for example, outputs:

```
<struct Train::Extras::CommandResult stdout="/etc/profile.d/lang.sh: line 19: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory\r\nbin dev  home  lib64\tmnt  proc  run\t shared  sys  usr      var\r\nboot  etc  lib\t media\topt  root  sbin  srv\t tmp  vagrant\r\n", stderr="", exit_status=0>
```

For comparison, this is the output _without_ the PTY:

```
<struct Train::Extras::CommandResult stdout="bin\nboot\ndev\netc\nhome\nlib\nlib64\nmedia\nmnt\nopt\nproc\nroot\nrun\nsbin\nshared\nsrv\nsys\ntmp\nusr\nvagrant\nvar\n", stderr="/etc/profile.d/lang.sh: line 19: warning: setlocale: LC_CTYPE: cannot change locale (UTF-8): No such file or directory\n", exit_status=0>
```

Note how `ls` behaves differently -- in the presence of a PTY, it will
prettify its output.

-----
Fixes #60.

This is still not a fix for _InSpec failing with the `requiretty` issue_, since the merging of `std{err,out}` will probably break most profiles and resources relying on the distinction. __However__, for users using train for remote execution, that do not that strongly depend on it, it's a necessary feature.

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -27,6 +27,7 @@ module Train::Transports
   #
   # @author Fletcher Nichol <fnichol@nichol.ca>
   class SSHFailed < Train::TransportError; end
+  class SSHPTYFailed < Train::TransportError; end
 
   # A Transport which uses the SSH protocol to execute commands and transfer
   # files.
@@ -55,6 +56,7 @@ module Train::Transports
     option :connection_retry_sleep, default: 1
     option :max_wait_until_ready, default: 600
     option :compression, default: false
+    option :pty, default: false
 
     option :compression_level do |opts|
       # on nil or false: set compression level to 0
@@ -96,6 +98,10 @@ module Train::Transports
 
       unless options[:password].nil?
         options[:auth_methods].push('password', 'keyboard-interactive')
+      end
+
+      if options[:pty]
+        logger.warn('[SSH] PTY requested: stderr will be merged into stdout')
       end
 
       super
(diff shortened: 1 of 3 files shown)


arlimus:
:+1:

PR #122, opened 2016-07-04
Use winrm v2 implementation

mwrock:
Test-Kitchen using winrm v2 has just been released. I'll be removing the WIP from kitchen-inspec momentarily as well.

### DIFF
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -3,7 +3,6 @@
 # author: Christoph Hartmann
 
 require 'base64'
-require 'winrm'
 require 'train/errors'
 
 module Train::Extras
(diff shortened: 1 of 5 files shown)


lib/train/extras/command_wrapper.rb
```diff
@@ -114,15 +113,22 @@ def run(script)
...
-    # reused from https://github.com/WinRb/WinRM/blob/master/lib/winrm/command_executor.rb
     # suppress the progress stream from leaking to stderr
     def safe_script(script)
       "$ProgressPreference='SilentlyContinue';" + script
     end
 
+    # Encodes the script so that it can be passed to the PowerShell
+    # --EncodedCommand argument.
+    # @return [String] The UTF-16LE base64 encoded script
+    def encoded(script)
```
chris-rock: +1, I like the approach to make this more explicit

lib/train/version.rb
```diff
@@ -3,5 +3,5 @@
 # Author:: Dominik Richter (<dominik.richter@gmail.com>)
 
 module Train
-  VERSION = '0.15.0'.freeze
+  VERSION = '0.16.0'.freeze
```
chris-rock: please lets not make a version bump within a PR. In train, we do a version bump, whenever we release.

appveyor.yml
```diff
@@ -10,7 +10,7 @@ environment:
   bundler_url: https://rubygems.org/downloads/bundler-1.9.9.gem
 
   matrix:
-    - ruby_version: "193"
+    - ruby_version: "21"
```
chris-rock: I assume winrm 2 is not supporting ruby 1.9.3 anymore?

chris-rock:
This looks awesome @mwrock Is this also addressing the issue, where we are only able to transfer scripts with a limited size?

lib/train/extras/command_wrapper.rb
mwrock: Yeah. Also, there is no `PowershellScript` class in winrm v2 so we cant leverage that.

appveyor.yml
mwrock: I actually just noticed this, but since 1.9.3 is so very obsolete and has long been unsupported, I don't think its worth changing winrm v2 to support it.

mwrock:
Thats right. There will be no script length limitation and according to local benchmarking, file uploads will be 15-30x faster. You can also run scripts within the same scope. So you could do:
```
shell.run("$blah = 'val'")
shell.run("write-host $blah")
```
Previously the second `run` call would have an entirely different `$blah`.

And winrmV2 can talk to and transfer files to Windows Nano server.

mwrock:
Created a separate PR for the travis/appveyor changes #123

mwrock:
No longer WIP :)

chris-rock:
💯 @mwrock

PR #123, opened 2016-07-04
Test ruby 2.1 instead of 1.9.3 and only launch one test group per travis/appveyor

mwrock:
Ruby 2.0 has been unsupported for some time now. Chef/Chef-DK ships with 2.1 and may soon bump to 2.2. This also gets rid of double travis/appveyor runs in PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,9 +14,13 @@ before_install:
   - gem install bundler
   - gem --version
 
+branches:
+  only:
+  - master
+
 matrix:
   include:
-  - rvm: 1.9.3
+  - rvm: 2.1
   - rvm: 2.0
   - rvm: 2.2
     script: bundle exec rake test test:docker config=test-travis-1.yaml
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ environment:
   bundler_url: https://rubygems.org/downloads/bundler-1.9.9.gem
 
   matrix:
-    - ruby_version: "193"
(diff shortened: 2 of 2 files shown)


arlimus:
I like the suggestion and reasoning @mwrock :+1: 

So far we have tried to keep older versions of Ruby supported, whenever reasonable. For example, users may still use an old yet supported OS release (RHEL 5) which doesn't support newer versions of Ruby. (of course we could package our own ruby, but that's separate from this standalone gem)

Is this by any chance related to https://github.com/chef/train/pull/122 in a way where Ruby <2.0 doesn't work with the update to WinRM? If so I think it's great to move forward :smile:

mwrock:
Yes @arlimus thats the motivation - the lack of compatibility with winrm v2. 1.9.3 was retired in 2/2015 https://www.ruby-lang.org/en/news/2015/02/23/support-for-ruby-1-9-3-has-ended/ so I think dropping this is fine.

arlimus:
Thank you Matt, this helps a lot!

Actually, ruby might have retired 1.9.3, but companies and users that require older yet still supported operating systems (like gen 5 of RHEL/CentOS; running until [31 March 2017](https://en.wikipedia.org/wiki/Red_Hat_Enterprise_Linux) ) may still rely on this version of Ruby. It is supported by the OS vendor. I have worked at companies that mandated this support, i.e. removing it may put these users in a difficult situation. I personally run latest and would always recommend that, but it's understandable that customers may have trouble with this.

PR #126, opened 2016-08-01
Bring train platform data more in line with ohai's platform data

stevendanna:
With this PR `family` refers to the distribuion family while `name` refers to the specific distribution. For example, the family attribute for ubuntu is `debian`. Further, architecture information is collected for Linux and OS X using `uname -m`.

This includes possibly breaking changes:

- The following platforms are now their own family and not listed as part of the `redhat` family: `amazon`, `fedora`.

- The following platforms are now part of the `redhat` family whereas before they were their own families: `centos oracle scientific enterpriseenterprise xenserver cloudlinux ibm_powerkvm nexus_centos wrlinux`

- The following platforms are now part of the `debian` family wheras before they were their own families: `ubuntu`, `linuxmint`, `raspbian`

- `opensuse` and `suse` are now part of a single `suse` family.

### DIFF
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -62,7 +62,7 @@ module Train::Extras
       },
     }
 
-    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware fedora} + OS['redhat'] + OS['debian'] + OS['suse']
+    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware fedora amazon} + OS['redhat'] + OS['debian'] + OS['suse']
 
     OS['unix'] = %w{unix aix hpux} + OS['linux'] + OS['solaris'] + OS['bsd']
 
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -38,15 +38,10 @@ module Train::Extras
     end
 
     OS = { # rubocop:disable Style/MutableConstant
-      'redhat' => %w{
-        redhat oracle centos fedora amazon scientific xenserver wrlinux
-      },
-      'debian' => %w{
-        debian ubuntu linuxmint raspbian
-      },
-      'suse' => %w{
-        suse opensuse
-      },
+      'redhat' => REDHAT_FAMILY,
+      'debian' => DEBIAN_FAMILY,
+      'suse' => SUSE_FAMILY,
+      'fedora' => %w{fedora},
       'bsd' => %w{
         freebsd netbsd openbsd darwin
       },
@@ -67,7 +62,7 @@ module Train::Extras
       },
     }
 
-    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware} + OS['redhat'] + OS['debian'] + OS['suse']
+    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware fedora} + OS['redhat'] + OS['debian'] + OS['suse']
 
     OS['unix'] = %w{unix aix hpux} + OS['linux'] + OS['solaris'] + OS['bsd']
 
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -22,6 +22,7 @@ module Train::Extras
     include Train::Extras::DetectWindows
     include Train::Extras::DetectEsx
 
+    attr_accessor :backend
     def initialize(backend, platform = nil)
       @backend = backend
       @platform = platform || {}
--- a/lib/train/extras/os_detect_darwin.rb
+++ b/lib/train/extras/os_detect_darwin.rb
@@ -7,9 +7,12 @@
 #   OHAI https://github.com/chef/ohai
 #   by Adam Jacob, Chef Software Inc
 #
+require 'train/extras/uname'
 
 module Train::Extras
   module DetectDarwin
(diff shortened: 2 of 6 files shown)


lib/train/extras/os_detect_linux.rb
```diff
@@ -16,91 +16,117 @@ module DetectLinux # rubocop:disable Metrics/ModuleLength
...
 
+    DEBIAN_FAMILY = %w{debian ubuntu linuxmint raspbian}.freeze
+    RHEL_FAMILY = %w{centos redhat oracle scientific enterpriseenterprise amazon xenserver cloudlinux ibm_powerkvm nexus_centos wrlinux}.freeze
+    SUSE_FAMILY = %w{suse opensuse}.freeze
+
+    def family_for_platform
+      if DEBIAN_FAMILY.include?(@platform[:name])
+        'debian'
+      elsif RHEL_FAMILY.include?(@platform[:name])
+        'rhel'
```
chris-rock: is there a reason, why we switch from `redhat`?

lib/train/extras/os_detect_linux.rb
```diff
@@ -16,91 +16,117 @@ module DetectLinux # rubocop:disable Metrics/ModuleLength
...
         @platform[:release] = raw.scan(/(\d+|\.+)/).join
       elsif !(raw = get_config('/etc/exherbo-release')).nil?
-        @platform[:family] = 'exherbo'
+        @platform[:name] = 'exherbo'
         # Because this is a rolling release distribution,
         # use the kernel release, ex. 4.1.6
         @platform[:release] = uname_r
       elsif !(raw = get_config('/etc/gentoo-release')).nil?
-        @platform[:family] = 'gentoo'
+        @platform[:name] = 'gentoo'
```
chris-rock: Should we add the family `gentoo` here as well? same applies for coreos and alpine

lib/train/extras/os_detect_linux.rb
```diff
@@ -16,91 +16,117 @@ module DetectLinux # rubocop:disable Metrics/ModuleLength
...
           @platform[:release] = os_info['VERSION']
         end
       end
 
+      @platform[:family] = family_for_platform
+
       !@platform[:family].nil? && !@platform[:release].nil?
     end
 
+    DEBIAN_FAMILY = %w{debian ubuntu linuxmint raspbian}.freeze
```
chris-rock: Do you think we should move constants to the top of the file?

lib/train/extras/os_detect_linux.rb
stevendanna: I'm happy to switch this back.  My goal in changing this was to bring it further in line with what `ohai` detects.

lib/train/extras/os_detect_linux.rb
stevendanna: Family is set to the name if no other name is set at line 110.

lib/train/extras/os_detect_darwin.rb
```diff
@@ -18,6 +18,9 @@ def detect_darwin
       # TODO: ditto on error
       return false if cmd.stdout.empty?
 
+      uname_m = @backend.run_command('uname -m').stdout.chomp
```
chris-rock: Do you think it may makes sense to into a method `detect_arch` for all detection implementations. This would allow us to know where it is located

lib/train/extras/os_detect_linux.rb
chris-rock: We need to be compatible with serverspec here: http://serverspec.org/advanced_tips.html They use redhat.

lib/train/extras/os_detect_linux.rb
chris-rock: Perfect. Missed that. That is great!

chris-rock:
Also see https://github.com/chef/chef/issues/4665 for a discussion around amazon linux

chris-rock:
fixes chef/inspec/issues/847

chris-rock:
fixes chef/inspec/issues/772

lib/train/extras/os_detect_linux.rb
```diff
@@ -12,95 +12,121 @@
 
 module Train::Extras
   module DetectLinux # rubocop:disable Metrics/ModuleLength
+    DEBIAN_FAMILY = %w{debian ubuntu linuxmint raspbian}.freeze
+    REDHAT_FAMILY = %w{centos redhat oracle scientific enterpriseenterprise amazon xenserver cloudlinux ibm_powerkvm nexus_centos wrlinux}.freeze
```
stevendanna: Per lamont's comment, we should drop amazon from this array

lib/train/extras/os_common.rb
```diff
@@ -66,7 +62,7 @@ def to_hash
       },
     }
 
-    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware} + OS['redhat'] + OS['debian'] + OS['suse']
+    OS['linux'] = %w{linux alpine arch coreos exherbo gentoo slackware fedora} + OS['redhat'] + OS['debian'] + OS['suse']
```
chris-rock: do we have to add amazon as a platform here?

arlimus:
Awesome improvements @stevendanna !! :+1:

chris-rock:
Awesome work @stevendanna !

PR #127, opened 2016-08-05
provide file_version and product_version for windows files

chris-rock:
fixes https://github.com/chef/inspec/issues/858

### DIFF
--- a/lib/train/extras/file_windows.rb
+++ b/lib/train/extras/file_windows.rb
@@ -69,11 +69,13 @@ module Train::Extras
     end
 
     def product_version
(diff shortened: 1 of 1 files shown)


stevendanna:
👍 seems reasonable.

chris-rock:
I had to remove the test for appveyor:

```
  it 'check file versioning on windows' do
    # may break when appveyor upgrades
    file = conn.file("C:/Ruby21/bin/ruby.exe")
    file.product_version.must_equal '...'
    file.file_version.must_equal '...'
  end
```

PR #130, opened 2016-08-16
Rescue EPIPE on connect in ssh transport

stevendanna:
SSH will raise an Errno::EPIPE if the remote server closes the
connection unexpectedly. This can happen, for example, in cases where a
user has an improperly configured ProxyCommand for that host in
`~/.ssh/config`.

Adding EPIPE to the set of exceptions to rescue ensures that we raise
Train::Transports::SSHFailed which callers of train my already be
rescuing.


chris-rock:
fixes https://github.com/chef/inspec/issues/840

chris-rock:
Thanks @stevendanna for this great improvement

PR #135, opened 2016-08-17
rescue docker erros during container shutdown

chris-rock:
This PR catches potential errors in our integration tests during docker shutdown. We can ignore errors during shutdown time, since they are not train related.


chris-rock:
This is not solving the issue.

PR #137, opened 2016-08-17
use --decode on base64 command to maintain compatibility with Darwin.

thomascate:
This is a fix for https://github.com/chef/train/issues/136. OSX was the only BSD I could find that was running an old enough version of base64 to have this issue. But this fix should catch any others.


chris-rock:
@thomascate great finding and great fix. 💯

PR #139, opened 2016-08-19
update integration test dependencies

tas50:


chris-rock:
Great improvements @tas50. Any reason why we remove 16.10?

tas50:
That was the part that failed previously

chris-rock:
@tas50 this is related to some docker issues we have in combination with travis. Those are timing errors and can be happen to all operating systems. We try to find the root cause.

PR #141, opened 2016-08-22
Add the darwin platform family

tas50:
Darwin is BSD, but it shouldn't be lumped in with BSD platforms. There's way too much different.


chris-rock:
@tas50 I was thinking about that as well. Great improvement.

PR #145, opened 2016-08-26
support ruby 2.2.1

chris-rock:


jkeiser:
You have my :+1: for what it's worth--that is indeed the actual cutoff point.

chris-rock:
Docker errors are not related to train

PR #146, opened 2016-08-31
use stat -c for alpine linux

chris-rock:
This fixes https://github.com/chef/inspec/issues/996


stevendanna:
👍 This looks reasonable to me.  If that conditional check for `-c` vs `--printf` gets any larger in the future we might consider extracting it to a method such as:

```
def has_stat_printf_flag(os)
```

But since it is only two platforms at the moment I think this is fine.

chris-rock:
@stevendanna agree with you on that.

Issue #148, opened 2016-09-07
Train doesn't create a login shell


jonathanmorley:
``` ruby
require 'train'

train = Train.create('ssh', host: '127.0.0.1', port: 2200, user: 'vagrant', password: 'vagrant')
conn = train.connection
puts conn.run_command("shopt -q login_shell && echo 'Login shell' || echo 'Not login shell'").stdout
```

Expected output `Login shell`
Actual output `Not login shell`

Since train 'logs in' to the remote host, and then runs commands against them, I would expect the resulting shells to be login shells.

I think a `login` parameter should be added to somehow force train to create a login shell (defaulted to false for backwards compatibility).

jonathanmorley:
The main driver for this issue is that inspec tests can return different results than manual ssh, since login shells will source the /etc/profile file.

chris-rock:
@jonathanmorley I really like that option. I see the point that paths etc may not be loaded. I just want to make sure we keep the default as is and make this fully optional. If users need this behavior, he should have the power to do so 👍

PR #149, opened 2016-09-08
Login shell

jonathanmorley:
Add `login` and `shell` options, to be able to force the command to run in a login shell

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -69,3 +69,5 @@ Style/SpaceAroundOperators:
   Enabled: false
 Style/IfUnlessModifier:
   Enabled: false
+Style/FrozenStringLiteralComment:
+  Enabled: false
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -96,10 +96,7 @@ module Train::Extras
 
       res = (@sudo_command || 'sudo') + ' '
 
-      unless @sudo_password.nil?
-        b64pw = Base64.strict_encode64(@sudo_password + "\n")
-        res = "echo #{b64pw} | base64 --decode | #{res}-S "
-      end
+      res = "#{safe_string(@sudo_password + "\n")} | #{res}-S " unless @sudo_password.nil?
 
       res << @sudo_options.to_s + ' ' unless @sudo_options.nil?
 
@@ -110,9 +107,18 @@ module Train::Extras
     # passed to the subshell
     def shell_wrap(cmd)
       return cmd unless @shell
+
       shell = @shell_command || '$SHELL'
       options = ' ' + @shell_options.to_s unless @shell_options.nil?
-      "echo '#{cmd}' > #{shell}#{options}"
+
+      "#{safe_string(cmd)} | #{shell}#{options}"
+    end
+
+    # encapsulates encoding the string into a safe form, and decoding for use.
+    # @return [String] A command line snippet that can be used as part of a pipeline.
+    def safe_string(str)
+      b64str = Base64.strict_encode64(str)
+      "echo #{b64str} | base64 --decode"
     end
   end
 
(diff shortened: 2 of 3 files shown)


jonathanmorley:
This is to address #148

chris-rock:
@jonathanmorley It a great idea to support the login shell. By default, I would expect that train would run in a  `non-interactive, non-login` mode. I would prefer to make this optional.

jonathanmorley:
This change keeps the default behaviour of train (the `shell` and `login` options default to false). With login set to true you get a login shell. I haven't looked into how this affects interectiveness

lib/train/extras/command_wrapper.rb
```diff
@@ -71,18 +74,22 @@ def verify
...
+        options[:login]
+      )
     end
 
     private
 
-    def build_prefix
-      return '' unless @sudo
-      return '' if @user == 'root'
+    def sudo_wrap(cmd)
```
chris-rock: Lets add comments what these function do

lib/train/extras/command_wrapper.rb
```diff
@@ -29,6 +29,8 @@ class LinuxCommand < CommandWrapperBase
     Train::Options.attach(self)
 
     option :sudo, default: false
+    option :shell, default: false
```
chris-rock: Its a little bit sub-optimal to switch types of a variable, I would prefer `nil` here, which would lead to the default

chris-rock: smme for login

chris-rock: Also lets add a comment to give users an idea of potential values

lib/train/extras/command_wrapper.rb
```diff
@@ -93,7 +100,15 @@ def build_prefix
...
       res << @sudo_options.to_s + ' ' unless @sudo_options.nil?
 
-      res
+      res + cmd
+    end
+
+    def shell_wrap(cmd)
+      return cmd unless @shell || @login
+      login_param = '--login ' if @login
+      shell = @shell.instance_of?(String) ? @shell : '$SHELL'
```
chris-rock: I would prefer not to check for the type here

arlimus:
Great idea :+1:

.rubocop.yml
```diff
@@ -69,3 +69,5 @@ Style/SpaceAroundOperators:
   Enabled: false
 Style/IfUnlessModifier:
   Enabled: false
+Style/FrozenStringLiteralComment:
```
chris-rock: Why are we changing the default setting?

chris-rock:
@jonathanmorley Thanks for incorporating my comments. Could you fix the functional test?
```
  1) Failure:
linux command::shell wrapping#test_0002_wraps sudo commands in a default shell with login [/home/travis/build/chef/train/test/unit/extras/command_wrapper_test.rb:73]:
--- expected
+++ actual
@@ -1 +1 @@
-"echo '0.7279336654738647' > $SHELL --login"
+"echo 'sudo 0.7279336654738647' > $SHELL --login"
```

.rubocop.yml
arlimus: I think it's because we don't yet have frozen string literal comments in our files:
http://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/FrozenStringLiteralComment

I think it's ok to disable for now and let's tackle this in a rubocop cleanup session :smile:

jonathanmorley:
Fixed. One remaining concern I have here is over quotes in the cmd. I have wrapped the cmd in single quotes so that the outer shell doesn't parse out any special characters, and through testing of things like

```
sh-3.2$ TEST1=test1 echo 'TEST2=test2; echo $TEST1; echo $TEST2; echo 'quotes'' | /bin/sh

test2
quotes
```

It seems to work as intended.

However, I see that for sudo we use base64 encoding. Any preferences towards the 2 approaches?

chris-rock:
Thanks for bringing up the question @jonathanmorley base64 proofed to be very reliant so far, therefore I would prefer that solution...

chris-rock:
💯 Thanks @jonathanmorley for this improvement! This travis errors are caused by docker timeouts not related to this PR

PR #151, opened 2016-09-20
Convert `wmic` architecture to a normal standard

jerryaldrichiii:
Windows `wmic os get` outputs values that are not expected for CPU architecture.

For example, instead of `x86_64` or `i386` it returns `64-bit` and `32-bit`

This uses the output of `wmic cpu get architecture` and converts it to something that is expected from commands like `uname` or `arch`.

This is intended to resolve the InSpec issue brought up by @seththoenen [here](https://github.com/chef/inspec/issues/1100) and is an alternative to my InSpec PR [here](https://github.com/chef/inspec/pull/1102).

### DIFF
--- a/lib/train/extras/os_detect_windows.rb
+++ b/lib/train/extras/os_detect_windows.rb
@@ -50,8 +50,37 @@ module Train::Extras
         @platform[:build] = sys_info[:BuildNumber]
         @platform[:name] = sys_info[:Caption]
         @platform[:name] = @platform[:name].gsub('Microsoft', '').strip unless @platform[:name].empty?
-        @platform[:arch] = sys_info[:OSArchitecture]
+        @platform[:arch] = read_wmic_cpu
       end
+    end
+
+    # `OSArchitecture` from `read_wmic` does not match a normal standard
+    # For example, `x86_64` shows as `64-bit`
+    def read_wmic_cpu
+      res = @backend.run_command('wmic cpu get architecture /format:list')
+      if res.exit_status == 0
+        sys_info = {}
+        res.stdout.lines.each { |line|
+          m = /^\s*([^=]*?)\s*=\s*(.*?)\s*$/.match(line)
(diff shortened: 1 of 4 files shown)


lib/train/extras/os_detect_windows.rb
```diff
@@ -7,6 +7,7 @@
 #   OHAI https://github.com/chef/ohai
 #   by Adam Jacob, Chef Software Inc
 #
+require 'pry'
```
chris-rock: please  remove pry

chris-rock:
Thanks @jerryaldrichiii this is a really good improvement.

lib/train/extras/os_detect_windows.rb
jerryaldrichiii: :facepalm: will do

chris-rock:
Works perfectly with Windows 2012 R2 and Windows 2016
screen shot 2016-09-20 at 9 50 25 pm
screen shot 2016-09-20 at 9 53 09 pm

chris-rock:
we missed to update the functional test for appveyor. please update the `test/windows/local` and `test/windows/winrm` with the following sections:

```
  it 'verify os' do
    os = conn.os
    os[:name].must_equal 'Windows Server 2012 R2 Datacenter'
    os[:family].must_equal 'windows'
    os[:release].must_equal '6.3.9600'
    os[:arch].must_equal 'x86_64'
  end
```
See https://github.com/chef/train/blob/master/test/windows/local_test.rb#L22-L28 and https://github.com/chef/train/blob/master/test/windows/winrm_test.rb#L28-L34

jerryaldrichiii:
Yup! I missed that too!

Tests have been updated.

chris-rock:
fixes https://github.com/chef/inspec/issues/1100

chris-rock:
Thanks @jerryaldrichiii the timeout error with docker is not related to this PR

PR #154, opened 2016-10-15
WIP support empty URIs

arlimus:
see a failures like [this](https://travis-ci.org/dev-sec/tests-ssh-hardening/jobs/167860903) which start appearing


arlimus:
awesome, it's failing! time to fix it :) I think we should continue to support these URIs for the simple reason that writing and printing it consistenly works so well, i.e. if we run with mock backend only, we print out

```
target: mock://
```

which users can use to configure the system again. Other solutions would involve setting the hostname, e.g.:

```
target: mock://dummy
```

which would be ignored (and may change in the future?). This convolutes the target with confusing information;

chris-rock:
+1

PR #155, opened 2016-10-16
Fix (most) VM integration tests

samcrang:
I noticed that VM integration tests invoked with `rake test:vm` didn't seem to pass. I spent a bit of time trying to fix these failures. I managed to fix most of the problems.

Unfortunately, I had some problems with CentOS 5.11 and Ubuntu 12.04:
- `bento/centos-5.11` using VirtualBox fails to start on my machine. I end up getting stuck in a "gurumeditation" state. This could be because of something weird with my machine.
- Centos 5.11 (i386) fails due to permissions on files in `/etc/sudoers.d`, possibly something to be fixed in the `sudo` cookbook?:

```
[vagrant@default-centos-511-i386 ~]$ sudo /bin/true
sudo: /etc/sudoers.d/reqtty is mode 0600, should be 0440
>>> /etc/sudoers.d/passwd: /etc/sudoers.d/reqtty near line 9 <<<
sudo: parse error in /etc/sudoers.d/passwd near line 9
sudo: no valid sudoers sources found, quitting
```
- The tests in `test/integration/sudo/passwd.rb` fail on Ubuntu 12.04.

I didn't want to spend too long looking at this because I wasn't sure if you actually run these tests anywhere? I'm assuming maybe you run them internally at Chef?

### DIFF
--- a/contrib/fixup_requiretty.rb
+++ b/contrib/fixup_requiretty.rb
@@ -14,11 +14,12 @@
 
 # FIXME
 user = ENV['TRAIN_SUDO_USER'] || 'todo-some-clever-default-maybe-current-user'
-sudoer = "/etc/sudoers.d/#{user}"
+sudo_conf_prefix = node['platform_family'] == 'freebsd' ? '/usr/local/etc' : '/etc'
+sudoer = "#{sudo_conf_prefix}/sudoers.d/#{user}"
 
 log "Warning: a sudoers configuration for user #{user} already exists, "\
     'doing nothing (override with TRAIN_SUDO_VERY_MUCH=yes)' do
-  only_if "test -f #{sudoer} || grep #{user} /etc/sudoers"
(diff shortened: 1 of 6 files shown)


chris-rock:
@samcrang Thanks for updating those. We moved most tests to kitchen-docker to be able to run those quickly with each PR. The VM-based tests are used mostly to cover systems that we can not run in docker eg. Windows, Mac, Solaris etc.

samcrang:
Ahhh, I see. Is there some trick to getting the Test Kitchen tests running with Docker? [I tried using the `kitchen-docker` driver](https://github.com/samcrang/train/commit/26691bde3bf36452baf51569ba406fd3b5885f21) but the tests fail because they expect a `vagrant` user.

jquick:
Due to the age of this and the new docker format I am closing this. If we are missing a critical platform we can see about adding it in.

PR #158, opened 2016-11-02
Add detection for Arista EOS

jerearista:
Note: EOS has 2 possible modes: the network CLI and a bash shell, typically only accessed in a test scenario such as with Vagrant.


lib/train/extras/os_detect_arista_eos.rb
```diff
@@ -0,0 +1,30 @@
...
+        @platform[:name] = 'arista_eos_bash'
+        family = 'fedora'
+      else
+        output = @backend.run_command('show version | json').stdout
+      end
+
+      unless output.empty?
+        eos_ver = JSON.parse(output)
+        @platform[:name] = @platform[:name] || 'arista_eos'
+        family ||= 'arista_eos'
```
chris-rock: Why are we not using `fedora` in this case?

jerearista: When you are in the bash shell, then you can do all the fedora things.  However, when you login the way you would to a  typical running switch, you are in a custom EOS CLI.  You can still prefix linux commands with 'bash' from this mode but common commands are things like 'show running-config', 'show interfaces status', etc.  Thus, things that expect to be in a regular linux shell will need special handling.

I'm definitely willing to go another route, if appropriate.

You can test yourself by registering at http://Arista.com/, then going to Software Downloads --> vEOS and downloading the files for your preferred hypervisor.

chris-rock: Maybe we should use `arista_eos` as the family for both instead of fedora?

jerearista: I assumed that more resources would adapt their behavior based on :family but since that's not the case, I'm fine with setting both to `arista_eos` if you think that's best.

jerearista: I was thinking about this a bit.   If we always mark EOS as family: arista_eos and include that as "linux", then resources will fail instead of skipping when we're in the usual CLI mode unless from the os detection I can specify to default to prefixing all linux commands with `bash` to drop to a linux shell (Ex: `bash uname -a` or `bash sudo cat /etc/passwd`).   

Alternatively, I could just patch up the resources as we make use of them.

chris-rock: Or we make `bash` default and use the EOS CLI only withinEOS specific InSpec resources. Is bash available on all devices?

chris-rock: We could achieve that by implementing a specific command wrapper for eos https://github.com/chef/train/blob/master/lib/train/extras/command_wrapper.rb#L164-L175

jerearista: I think that will work.   I'm not sure I'll get the wrapper in by Friday but I don't think that needs to hold up the initial platform support.  So, to recap, I'll set the family to `arista_eos` for both cases and we can improve handling the differences with a command wrapper.

chris-rock:
Thanks @jerearista Lets cover the other improvements in another PR

PR #159, opened 2016-11-03
Add openvms detection

briandoodyie:
This has been tested with OpenVMS 8.3.

* Added openvms detection
* Removed the hardcoding of ssh_connection setting platform to unix. 
* In os_common.rb we can now more accurately determine the platform by using uname_s method. Openvms does not have understand uname_s and doesnt generate an error detectable by the command object. However you can check the message coming from that for the generic openvms error.

### DIFF
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -99,7 +99,7 @@ module Train::Extras
       # to Windows and others
       case uname_s
       when /unrecognized command verb/
-        @platform[:family] = 'openvms'  
+        @platform[:family] = 'openvms'
       when /linux/i
         @platform[:family] = 'linux'
       when /./
@@ -120,7 +120,7 @@ module Train::Extras
       return detect_windows if pf == 'windows'
       return detect_darwin if pf == 'darwin'
       return detect_esx if pf == 'esx'
-      return detect_openvms if pf =="openvms"
+      return detect_openvms if pf =='openvms'
 
       if %w{freebsd netbsd openbsd aix solaris2 hpux}.include?(pf)
         return detect_via_uname
@@ -129,7 +129,7 @@ module Train::Extras
       # unix based systems combine the above
       return true if pf == 'unix' and detect_darwin
       return true if pf == 'unix' and detect_esx
-      #This is assuming that pf is set to unix, this should be if pf == 'linux'
+      # This is assuming that pf is set to unix, this should be if pf == 'linux'
       return true if pf == 'unix' and detect_arista_eos
       return true if pf == 'unix' and detect_via_uname
 
--- a/lib/train/extras/os_common.rb
+++ b/lib/train/extras/os_common.rb
@@ -14,6 +14,7 @@ require 'train/extras/os_detect_unix'
 require 'train/extras/os_detect_windows'
 require 'train/extras/os_detect_esx'
 require 'train/extras/os_detect_arista_eos'
+require 'train/extras/os_detect_openvms'
 
 module Train::Extras
   class OSCommon
@@ -23,6 +24,7 @@ module Train::Extras
     include Train::Extras::DetectWindows
     include Train::Extras::DetectEsx
     include Train::Extras::DetectAristaEos
+    include Train::Extras::DetectOpenVMS
 
     attr_accessor :backend
     def initialize(backend, platform = nil)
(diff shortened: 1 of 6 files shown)


lib/train/extras/os_common.rb
```diff
@@ -96,6 +98,8 @@ def detect_family
       # TODO: extend base implementation for detecting the family type
       # to Windows and others
       case uname_s
+      when /unrecognized command verb/
```
chris-rock: not sure if we can expect that every system that has not uname_s is openvms

chris-rock: maybe we just don't set any family then

briandoodyie: problem is that if we omit this part, it matches 

 when /./
        @platform[:family] = 'unix'

briandoodyie: Its a pain because the uname_s command doesnt returns an error that can be detected by the command object (old os) but a message that says : **%DCL-W-IVVERB, unrecognized command verb - check validity and spelling**

lib/train/extras/os_common.rb
```diff
@@ -124,6 +129,7 @@ def detect_family_type # rubocop:disable Metrics/CyclomaticComplexity, Metrics/P
       # unix based systems combine the above
       return true if pf == 'unix' and detect_darwin
       return true if pf == 'unix' and detect_esx
+      #This is assuming that pf is set to unix, this should be if pf == 'linux'
```
chris-rock: I think we can argue that arista should be in detect_linux

lib/train/extras/os_common.rb
chris-rock: I think we should create a method for detecting non-uname operating systems. Is there a reliable command that only works on openvms to detect the family?

chris-rock: Okay, lets use this for now until we find any other `non-uname` operating system

briandoodyie:
Looking at the lint testing part now

briandoodyie:
@chris-rock not sure on this final error. All the lint testing should pass now.

lib/train/extras/os_common.rb
```diff
@@ -124,6 +129,7 @@ def detect_family_type # rubocop:disable Metrics/CyclomaticComplexity, Metrics/P
       # unix based systems combine the above
       return true if pf == 'unix' and detect_darwin
       return true if pf == 'unix' and detect_esx
+      # This is assuming that pf is set to unix, this should be if pf == 'linux'
```
chris-rock: we should remove that command, since from trains perspective linux is a subset of a unix-like operating system.

lib/train/extras/os_common.rb
```diff
@@ -96,6 +98,8 @@ def detect_family
       # TODO: extend base implementation for detecting the family type
       # to Windows and others
       case uname_s
+      when /unrecognized command verb/
+        @platform[:family] = 'openvms'
```
chris-rock: Since we cannot expect that every non-uname system is openvms I would prefer to set it to nil here?

lib/train/transports/ssh_connection.rb
```diff
@@ -230,7 +230,7 @@ def to_s
 
     class OS < OSCommon
       def initialize(backend)
-        super(backend, { family: 'unix' })
```
chris-rock: I just thinking if it is not easier to count openvms as a unix-like system as well? Not sure about that. This would allow us to just use 

```
return true if pf =='unix' && detect_openvms 
```

chris-rock:
awesome @briandoodyie I think we are getting very close...

chris-rock:
@briandoodyie Anything I can help with?

lib/train/transports/ssh_connection.rb
briandoodyie: This was the original code and the problem there is that

briandoodyie:
Hmmm the ssh access will work and was the original change that was made. It made everything a lot simpler but doesnt windows now also support ssh. Might be assumption that just doesnt work now. thoughts? @chris-rock

chris-rock:
@briandoodyie I wanna clean this up as well, I just want to do this in two steps.

chris-rock:
@briandoodyie I go with the current PR. Thanks @briandoodyie for putting all the effort into that

PR #167, opened 2016-12-21
Fix Oracle Linux detection

carldjohnston:
get_config path missing '/etc/' prefix


chris-rock:
Awesome. Thank you @carldjohnston

PR #172, opened 2017-03-24
Fix Net::SSH warning when passing nil option values

tylercloke:
Net::SSH now complains if you pass in values that are nil as part of the option hash. Delete the nil values before passing to Net::SSH.


tylercloke:
Thanks for the fast feedback @chris-rock and @adamleff!

PR #176, opened 2017-05-19
Fixed IPv6 URI parsing

zfjagann:
Currently if you pass a URI with a literal IPv6 address in it, it doesn't unwrap the square brackets around the IP. This leads to failures downstream when this is passed to `getaddrinfo`, which requires the IP without the brackets.

This PR just strips out the square brackets around the host parameter when parsing it from the URI by calling `URI.hostname` instead of `URI.host`.

Also added some unit tests for this case, and the IPv4 case.

### DIFF
--- a/lib/train.rb
+++ b/lib/train.rb
@@ -61,7 +61,7 @@ module Train
     uri = parse_uri(conf[:target].to_s)
     unless uri.host.nil? and uri.scheme.nil?
       conf[:backend]  ||= uri.scheme
-      conf[:host]     ||= uri.host
+      conf[:host]     ||= uri.hostname
       conf[:port]     ||= uri.port
       conf[:user]     ||= uri.user
       conf[:password] ||= uri.password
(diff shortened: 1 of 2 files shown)


chris-rock:
Using a method that is not well documented in Ruby. Nice job @zfjagann I found the reference for hostname in https://github.com/ruby/ruby/blob/trunk/lib/uri/generic.rb#L234-L240

```
inspec> u = URI 'mock://1.2.3.4:123'
=> #<URI::Generic mock://1.2.3.4:123>
inspec> u.host
=> "1.2.3.4"
inspec> u.hostname
=> "1.2.3.4"
inspec> u = URI 'mock://[abc::def]:123'
=> #<URI::Generic mock://[abc::def]:123>
inspec> u.host
=> "[abc::def]"
inspec> u.hostname
=> "abc::def"
```

chris-rock:
@zfjagann I also want to apologize for the late response. ChefConf and travel kept us very busy :-)

PR #178, opened 2017-05-25
Remove autoloads in favor of eager loading

Sharpie:
The Ruby Autoload feature has been deprecated as it is not thread safe.  This
patch replaces the use of autload with simple requires.

Ref: http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/41149

### DIFF
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -12,14 +12,14 @@ module Train::Transports
 
     include_options Train::Extras::CommandWrapper
 
-    autoload :File, 'train/transports/local_file'
-    autoload :OS,   'train/transports/local_os'
-
     def connection(_ = nil)
       @connection ||= Connection.new(@options)
     end
 
     class Connection < BaseConnection
+      require 'train/transports/local_file'
+      require 'train/transports/local_os'
+
       def initialize(options)
         super(options)
         @cmd_wrapper = nil
--- a/lib/train.rb
+++ b/lib/train.rb
@@ -3,6 +3,7 @@
 # Author:: Dominik Richter (<dominik.richter@gmail.com>)
 
 require 'train/version'
+require 'train/options'
 require 'train/plugins'
 require 'train/errors'
 require 'uri'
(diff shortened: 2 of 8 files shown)


Sharpie:
Without this change, parallel operations over train connections would often fail with missing contant errors:

```
uninitialized constant Train::Transports::SSH::Connection::OS
```

This matches the behavior of how autoload clobbers constants in a thread-unsafe manner as described here:

https://prezi.com/ff9yptxhohjz/making-autoload-thread-safe/

Sharpie:
Another explanation of what happens to constants during autoload:

http://blade.nagaokaut.ac.jp/cgi-bin/scat.rb/ruby/ruby-core/20238

Sharpie:
Looking at the Travis build history, it looks like docker tests often get stuck and time out.

lib/train/transports/local.rb
```diff
@@ -12,14 +12,14 @@ class Local < Train.plugin(1)
...
 
-    autoload :File, 'train/transports/local_file'
-    autoload :OS,   'train/transports/local_os'
-
     def connection(_ = nil)
       @connection ||= Connection.new(@options)
     end
 
     class Connection < BaseConnection
+      require 'train/transports/local_file'
```
chris-rock: I think we should require where we had the autoload

chris-rock:
This refers to https://bugs.ruby-lang.org/issues/5653

arlimus:
@Sharpie Sorry for the late response, we just came back from extended travel.
Awesome contribution, would love to get it in! What do you think about Chris' comment?

Sharpie:
I think there may have been a reason I moved that require... perhaps the `BaseConnection` class had to exist before local_file could define additional classes underneath it. I'll do some testing this evening and confirm whether that was the case.

lib/train/transports/local.rb
Sharpie: Looks like the require had to be here as `local_file` starts out by declaring a class underneath `Connection`:

    class Train::Transports::Local::Connection
      class File < LinuxFile
        ...

PR #183, opened 2017-06-27
Fix to allow kitchen verify to pass with ssh transport

jalberti:
By default keys_only is set to true, this results in kitchen verify to fail unless a password or key file is explicitly set in .kitchen.yml, we can fix this by ensuring that not only publickey is pushed into the auth_methods, but also keys_only is set to false.

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -102,6 +102,7 @@ module Train::Transports
         else
           logger.debug('[SSH] Using Agent keys as no password or key file have been specified')
           options[:auth_methods].push('publickey')
+          options[:keys_only] = false # set keys_only to false to ensure agent keys are actually used
         end
       end
 


chris-rock:
@jalberti is that something we want to cover in kitchen-inspec? https://github.com/chef/kitchen-inspec/blob/master/lib/kitchen/verifier/inspec.rb#L197-L221

jalberti:
@chris-rock I'm a little pressed by time, and not an inspec expert by any means, can you drop me the cli command that allows me to run the test, then I can see if I'm able to quickly implement a breaking test that would be fixed by my change.

adamleff:
@jalberti Hi, and thanks for submitting this issue! I'm going to take a look at it for you.

I was able to reproduce this by doing the following:
 * rename my keys to non-standard file names so that Test Kitchen didn't pick them up automatically
 * Delete all identities from my SSH Agent
 * add my AWS SSH key to my SSH Agent
 * wrote a .kitchen.yml that uses a CentOS 7 image with the only transport setting being the username set to `centos`
 * ran `kitchen verify`

The create and the converge step worked perfectly and used my SSH agent credentials. The verify step failed with:

```
>>>>>> ------Exception-------
>>>>>> Class: Kitchen::ActionFailed
>>>>>> Message: 1 actions failed.
>>>>>>     Failed to complete #verify action: [Transport error, can't connect to 'ssh' backend: SSH session could not be established] on default-centos
>>>>>> ----------------------
```

However, if I ran `inspec shell --target ssh://centos@ec2-x-x-x-x.compute-1.amazonaws.com`, it succeeded!

Therefore, I agree with @chris-rock that we likely need to make a change to kitchen-inspec instead. InSpec itself seems to be doing the right thing. The change will likely be to not set `keys_only` at all in kitchen-inspec and just let InSpec do the right thing.

I have some more testing to do, but I suspect I'll have a PR submitted to kitchen-inspec this afternoon after I verify. Thanks again for submitting the issue and I'll get back to you soon.

adamleff:
Hi again, @jalberti!

I've opened up https://github.com/chef/kitchen-inspec/pull/140 which will address this in kitchen-inspec. I confirmed Train already does the right thing and only sets keys_only to true if keys are provided, and in the case where key files are not provided, it checks the running Agent process to see if identities are available and does not set the parameter.

I'd like to leave this logic in Train the way it is because it eliminates any blocking by net-ssh prompting a user for a password in the case where they provide a key and the key is not authorized.

Please feel free to subscribe to the kitchen-inspec PR for updates. As soon as we can get that merged, I'm sure we can cut a release in short order.

Thanks again! I'm closing this PR now, but please reach out if we can help or if we need to re-open.

~Adam

jalberti:
@adamleff sounds good, thx

PR #196, opened 2017-09-16
[win-perf][proof-of-concept] run PS in a server locally

sdelano:
## Proof of Concept - Do Not Merge

we incur a significant overhead when running windows commands
locally because each command needs to execute within a new
powershell instance. this adds an overhead of approximately
250ms per powershell request.

the changes in the sd/win-perf branch of inspec go a ways
towards reducing the number of powershell commands that are
executed, but they will never overcome the overhead that's
baked in for each one of those calls.

this change is a PROOF OF CONCEPT for what it would look
like to have a long-running powershell process that inspec/train
can send commands to. in this POC, it's implemented as an HTTP
server written in powershell that executed the body of POST
requests and has a special `exit` command to kill the process.

on `master` of inspec, this reduces the execution time of the
`windows-baseline` profile from 30s to 13s. when combined with
the `sd/win-perf` branch of inspec, the execution time is
further reduced to 9s.

Signed-off-by: Stephen Delano <stephen@chef.io>

### DIFF
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -6,6 +6,9 @@
 require 'train/plugins'
 require 'mixlib/shellout'
 
+require 'win32/process'
+require 'base64'
+
 module Train::Transports
   class Local < Train.plugin(1)
     name 'local'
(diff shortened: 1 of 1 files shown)


arlimus:
Hi @sdelano great numbers through this improvement, I love how the total execution goes down!! 👍 

A bit concerned with the method though, I don't think holding the process + open port will be something we can broadly support, especially with security guidelines on most systems (i.e. don't open any ports that can be exploited).

@chris-rock @adamleff opinions?

chris-rock:
That is a great way, but I agree with Dom. I think we also could try to re-use the same powershell session and run multiple powershell commands via the same session. Maybe @mwrock has some ideas?

sdelano:
@arlimus @chris-rock I 100% agree that opening an port for HTTP is not the way to go. I chose HTTP here because it was quick and easy in PowerShell and it proved that the method worked.

I've been working this week on taking the same approach and using IPC via pipes for the same effect: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365574%28v=vs.85%29.aspx?f=255&MSPPError=-2147217396#base.using_pipes_for_ipc

This is very similar to what we're currently using for erlang -> ruby communication via the erlectricity (github.com/chef/erlectricity) project.

jerryaldrichiii:
Closing this since work was implemented via https://github.com/chef/train/commit/d5dbab364a543070a04738a33e7866e116e2c99d

PR #199, opened 2017-09-25
Continue to support older net-ssh while fixing 4.2 deprecation

adamleff:
In #197, we bumped our dependency on net-ssh to ~> 4.2 to address a deprecation of the `paranoid` option made in net-ssh 4.2.0. Unfortunately, this is causing users of newer InSpec versions to not be able to use InSpec with Chef v12 due to net-ssh v3 in use in that version.

This change reverts the net-ssh pin to be looser and dynamically chooses the connection option to use depending on the version of net-ssh in use. This will avoid the deprecation warning and allow us to support newer net-ssh versions.

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -33,7 +33,7 @@ module Train::Transports
   # files.
   #
   # @author Fletcher Nichol <fnichol@nichol.ca>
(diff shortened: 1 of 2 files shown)


kplimack:
🥇

PR #201, opened 2017-10-06
implement uploads and downloads for ssh and winrm

thommay:
also, scp uploads are now parallelisable.


adamleff:
@thommay looks good, but you've got some angry tests (catching an actual problem... who thought Rubocop could actually do that?)

Lemme know when fixed and I'll take another look.

thommay:
thanks @adamleff . done

PR #205, opened 2017-11-06
Kubernetes transport

coderanger:
Includes a test harness using the same tests as for Docker.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,16 @@ matrix:
   - rvm: 2.4.1
     script: bundle exec rake $SUITE
     env: SUITE="test:docker config=test-travis-ubuntu.yml"
+  - rvm: 2.4.1
+    env:
+    - CHANGE_MINIKUBE_NONE_USER=true
+    script:
+    - curl -Lo kubectl https://storage.googleapis.com/kubernetes-release/release/v1.7.0/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin
+    - curl -Lo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin
+    - sudo minikube start --vm-driver=none --kubernetes-version=v1.8.0
+    - minikube update-context
+    - JSONPATH='{range .items[*]}{@.metadata.name}:{range @.status.conditions[*]}{@.type}={@.status};{end}{end}'; until kubectl get nodes -o jsonpath="$JSONPATH" 2>&1 | grep -q "Ready=True"; do sleep 1; done
+    - bundle exec rake test:kubernetes
 
 deploy:
   provider: rubygems
--- a/lib/train/transports/kubernetes.rb
+++ b/lib/train/transports/kubernetes.rb
@@ -1,0 +1,73 @@
+# encoding: utf-8
+#
+# Copyright 2017, Noah Kantrowitz
+
+require 'mixlib/shellout'
+
+require 'train/errors'
+
+module Train::Transports
+  class SSHFailed < Train::TransportError; end
+
+  class Kubernetes < Train.plugin(1)
+    name 'kubernetes'
+
+    include_options Train::Extras::CommandWrapper
+    option :pod, required: true
+    option :container, default: nil
+    option :kubectl_path, default: 'kubectl'
+
+    def connection(state = {})
+      opts = merge_options(options, state || {})
+      validate_options(opts)
+      opts[:logger] ||= logger
+      unless @connection && @connection_opts == opts
+        @connection ||= Connection.new(opts)
+        @connection_opts = opts.dup
+      end
+      @connection
+    end
+
+    class Connection < BaseConnection
+      def os
+        @os ||= OS.new(self)
+      end
+
+      def file(path)
+        @files[path] ||= Train::File::Remote::Linux.new(self, path)
+      end
+
+      def run_command(cmd)
+        kubectl_cmd = [options[:kubectl_path], 'exec']
+        kubectl_cmd.concat(['--container', options[:container]]) if options[:container]
+        kubectl_cmd.concat([options[:pod], '--', '/bin/sh', '-c', cmd])
+
+        so = Mixlib::ShellOut.new(kubectl_cmd, logger: logger)
+        so.run_command
+        if so.error?
+          # Trim the "command terminated with exit code N" line from the end
+          # of the stderr content.
+          so.stderr.gsub!(/command terminated with exit code #{so.exitstatus}\n\Z/, '')
+        end
+        CommandResult.new(so.stdout, so.stderr, so.exitstatus)
+      end
+
+      def uri
+        if options[:container]
+          "kubernetes://#{options[:pod]}/#{options[:container]}"
+        else
+          "kubernetes://#{options[:pod]}"
+        end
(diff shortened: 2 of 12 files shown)


adamleff:
Thanks for the contribution, @coderanger! We're discussing this one internally in relation to some other kubernetes work we've been discussing. Hopefully we'll have more feedback for you next week.

.travis.yml
```diff
@@ -40,6 +40,16 @@ matrix:
   - rvm: 2.4.1
     script: bundle exec rake $SUITE
     env: SUITE="test:docker config=test-travis-ubuntu.yml"
+  - rvm: 2.4.1
```
chris-rock: Nice to have the integration tests included. That is great!

test/integration/kubernetes_test.rb
```diff
@@ -0,0 +1,152 @@
+# encoding: utf-8
```
chris-rock: Can we move that file to `test/integration/k8s/test.rb`?

test/integration/kubernetes_test.rb
```diff
@@ -0,0 +1,152 @@
...
+#
+# Copyright 2017, Noah Kantrowitz
```
chris-rock: Please change that to Author.

test/integration/helper.rb
```diff
@@ -31,7 +31,7 @@ def root_group(os)
     end
 
     def selinux_label(backend, path = nil)
-      return nil if backend.class.to_s =~ /docker/i
+      return nil if backend.class.to_s =~ /docker|kubernetes/i
```
chris-rock: We are preparing https://github.com/chef/inspec/issues/1661 in Train and InSpec. kubernetes will become a platform in train. At this point, we are adding a k8s transport for the execution inside of containers only. We think `k8s-exec` is a better name for this specific implementation.

lib/train/transports/kubernetes.rb
```diff
@@ -0,0 +1,72 @@
...
+      def os
+        @os ||= OS.new(self)
+      end
+
+      def file(path)
+        @files[path] ||= Train::File::Remote::Linux.new(self, path)
+      end
+
+      def run_command(cmd)
+        kubectl_cmd = [options[:kubectl_path], 'exec']
```
chris-rock: This implementation relies on shelling out to kubectl rather than using the API directly. This limits the usage and decreases the usability of train/inspec. Is there any reason why you decided against using https://github.com/abonas/kubeclient  directly?

lib/train/transports/kubernetes.rb
```diff
@@ -0,0 +1,72 @@
+# encoding: utf-8
+#
+# Copyright 2017, Noah Kantrowitz
```
chris-rock: Please change to author

test/integration/tests/path_folder_test.rb
```diff
@@ -76,7 +72,11 @@
 
     it 'has selinux label handling' do
       res = Test.selinux_label(backend, file.path)
-      file.selinux_label.must_equal(res)
+      if res.nil?
```
chris-rock: Why do you expect that is a result from a method is nil, that this is okay?

lib/train/transports/kubernetes.rb
coderanger: Copyright is correct as I remain the owner of this code. Unless you're requesting I assign copyright to Chef Software, which would be unusual?

lib/train/transports/kubernetes.rb
coderanger: A couple of reasons:

1. Stability, the APIs change a lot and kubectl/client-go will always be the first class citizen in the ecosystem.
2. Performance, Ruby network perf is not great and because we're using it as the file transport system too it would be annoying.
3. Support, Kubeclient (and pretty much all other client libs) don't actually implement the `exec` API. Most libraries just do the REST API, which here wouldn't actually help with anything. The `exec` API is weirdly undocumented but it initiates over an HTTP POST and then upgrades to SPDY (which will be replaced with websockets soon). Implementing a client for this in Ruby would be very annoying.

test/integration/helper.rb
coderanger: I don't understand, that PR says "platform" would be things like RedHat or Ubuntu, in which case Kubernetes will not be a platform. The OS image used in the underlying container would be the platform, just like with Docker.

test/integration/kubernetes_test.rb
coderanger: Sure, this was used to match with the existing Docker tests.

test/integration/tests/path_folder_test.rb
coderanger: As noted above, you already returned `nil` here during the Docker tests, and I extended that to also return nil on Kubernetes tests. You weren't actually handling that in most of the tests so I fixed that while making the Kubernetes tests pass.

aaronlippold:
How would this connect to openShift as well? Would we want a individual openshift or oc:// transport or would it be a "type" of k8r transport do you think?

frezbo:
@aaronlippold as OC still uses the same k8s API server, it think it would be ok to use the same transport that k8s uses. Maybe in inspec we could alias k8s transport to oc.

chris-rock:
I agree with @frezbo here, we should keep it to plain k8s. I also do not recommend to use an alias since this may confuses the community. The confusion may come from the fact that this implementation uses kubectl. I would like to avoid that and instead talk to the k8s api directly. Then we are talking about API compatible solutions not binary compatible solutions

frezbo:
+1 on the usage of API. I hope we use this: https://github.com/kubernetes-client/ruby seems there are a lot of k8s ruby API's

chris-rock:
That is a great reference @frezbo I missed that we have near official clients for ruby now

coderanger:
No, it's not a great a reference, that's an abandoned project from 8 months ago. And it doesn't expose the pod execution API because almost no clients do. I explained why I used `kubectl` last year and all the same reasons apply.

frezbo:
That's sad. I really hope all other API's gets the same love as go-client.

coderanger:
Maybe in a long long time when the API stabilizes more fully. Until then, that seems highly unlikely.

coderanger:
As for the original question, the `oc` command is mostly just an extension to `kubectl` so to make that work we would just need to set `kubectl_path: 'oc'`.

clintoncwolfe:
We will one day get around to adding a kubernetes transport; it will likely be a separate plugin, however, and may be "chimeric" (internally using multiple means to answer questions - perhaps both kubectl and some ruby api (a cursory glance indicates that https://github.com/kubernetes-client/ruby appears to be active as of right now)).  

This is excellent work, and when a kube plugin is started in earnest, we'll certainly refer to this (with attribution and copyright if we're outright copying).

For now though, this PR isn't likely to ever be merged into train (and train has 18 months of growth since this was last touched) so, I'm going to close this.  It's linked from the issue, so it won't get misplaced.

bgeesaman:
https://github.com/bgeesaman/inspec-k8s which relies on https://github.com/bgeesaman/train-kubernetes was recently released.  `train-k8s` was already reserved on rubygems, so `train-kubernetes` attempts to permit assertions on Kubernetes API objects and their specs (the "k8s" transport mentioned by @clintoncwolfe above).  It does not implement `kubectl exec` like functionality like this PR does (the "k8s-container" transport mentioned above).  Feedback is welcomed.

PR #209, opened 2017-11-13
Platform framework and detect DSL

jquick:
This change moves the detect logic for platforms into a new framework. You can create platforms on the fly using the new platform DSL.


lib/train/transports/mock.rb
```diff
@@ -77,8 +74,20 @@ def mock_os(value = {})
       # if a user passes a nil value, set to an empty hash so the merge still succeeds
       value ||= {}
 
-      os_params = { name: 'unknown', family: 'unknown', release: 'unknown', arch: 'unknown' }.merge(value)
```
adamleff: Hey @jquick! We need to ensure `mock_os` still allows other callers to set the release and arch, and we also need to make sure they're not nil. Can you check to see if you account for this, and if not, could you push up a commit that adds that functionality back? 

This will potentially break `only_if` evaluation during `inspec check` calls as well as some unit tests potentially (though I think everything in InSpec's unit tests are focused on name and family only).

PR #214, opened 2017-11-17
Enable caching on connections

jquick:
This changes adds a new connection called "CacheConnection". When passed a connection it will cache file and command calls. All connection methods will be accessible on the cached version.


lib/train/plugins/base_connection.rb
```diff
@@ -52,31 +58,44 @@ def local?
...
-
     # we need to keep os as a method for backwards compatibility with inspec
     alias os platform
 
+    # Execute a command using this connection.
+    #
+    # @param command [String] command string to execute
+    # @return [CommandResult] contains the result of running the command
+    def run_command_via_connection(_command)
+      fail Train::ClientError, "#{self.class} does not implement #run_command_via_connection()"
```
adamleff: I think we should raise a `NotImplementedError` instead as that's the standard exception class for this type of issue. Is there a reason we shouldn't?

lib/train/plugins/base_connection.rb
```diff
@@ -52,31 +58,44 @@ def local?
...
+
     # Interact with files on the target. Read, write, and get metadata
     # from files via the transport.
     #
     # @param [String] path which is being inspected
     # @return [FileCommon] file object that allows for interaction
-    def file(_path, *_args)
-      fail Train::ClientError, "#{self.class} does not implement #file(...)"
+    def file_via_connection(_path, *_args)
+      fail Train::ClientError, "#{self.class} does not implement #file_via_connection(...)"
```
adamleff: I think we should raise a `NotImplementedError` instead as that's the standard exception class for this type of issue. Is there a reason we shouldn't?

lib/train/plugins/cache_connection.rb
```diff
@@ -0,0 +1,45 @@
...
+      @cache = {}
+
+      # default caching options
+      @cache_enabled = {
+        file: true,
+        command: false,
+      }
+
+      @cache_enabled.each_key do |type|
+        @cache[type] = {}
```
adamleff: Maybe call `clear_cache` here instead?

lib/train/plugins/cache_connection.rb
```diff
@@ -0,0 +1,45 @@
+# encoding: utf-8
+
+class Train::Plugins::Transport
+  class CacheConnection
+    attr_accessor :cache_enabled
```
adamleff: Ruby nitpick, I'd rather see a `cache_enabled?` method that looks at this hash and returns true/false... just to keep consistency with `?` methods returning booleans. If you choose to take this advice, this will obviously require some changes in BaseConnection, but not too many.

lib/train/plugins/cache_connection.rb
```diff
@@ -0,0 +1,45 @@
...
+        @cache[type] = {}
+      end
+    end
+
+    def clear_cache(type)
+      @cache[type.to_sym] = {}
+    end
+
+    def file(path)
+      if @cache[:file].key?(path)
```
adamleff: Could this be simplified to?:

```
@cache[:file][path] ||= @connection.file_via_connection(path)
```

That should return the value if it's there, otherwise, assign the new value and then return it.

lib/train/plugins/cache_connection.rb
```diff
@@ -0,0 +1,45 @@
...
+        @cache[:file][path]
+      else
+        @cache[:file][path] = @connection.file_via_connection(path)
+      end
+    end
+
+    def run_command(cmd)
+      if @cache[:command].key?(cmd)
```
adamleff: Same as above - I think we can do `||=` and simplify a bit.

test/unit/plugins/connection_test.rb
```diff
@@ -14,6 +14,10 @@
       proc { connection.run_command('') }.must_raise Train::ClientError
     end
 
+    it 'provides a run_command_via_connection method' do
```
adamleff: I wonder if this should be `it 'raises an exception for run_command_via_connection'`, and then we add a `it 'provides a run_command_via_connection method'` check for each of our connection classes. Thoughts?

test/unit/plugins/connection_test.rb
```diff
@@ -22,6 +26,10 @@
       proc { connection.file('') }.must_raise Train::ClientError
     end
 
+    it 'provides a file_via_connection method' do
+      proc { connection.file_via_connection('') }.must_raise Train::ClientError
```
adamleff: Same as above...

jquick:
This should bet updated with all changes to date.

lib/train/plugins/base_connection.rb
```diff
@@ -52,39 +72,54 @@ def local?
...
     def login_command
-      fail Train::ClientError, "#{self.class} does not implement #login_command()"
+      fail NotImplementedError, "#{self.class} does not implement #login_command()"
+    end
+
+    # Execute a command using this connection.
+    #
+    # @param command [String] command string to execute
+    # @return [CommandResult] contains the result of running the command
+    def run_command_via_connection(_command)
```
chris-rock: Those two methods need to be private

lib/train/transports/docker.rb
```diff
@@ -69,18 +69,27 @@ def close
...
-    def file(path)
-      @files[path] ||=\
-        if os.aix?
-          Train::File::Remote::Aix.new(self, path)
-        elsif os.solaris?
-          Train::File::Remote::Unix.new(self, path)
-        else
-          Train::File::Remote::Linux.new(self, path)
-        end
+    def uri
```
chris-rock: Is there a reason to make this public?

lib/train/transports/local.rb
```diff
@@ -23,34 +23,35 @@ def initialize(options)
...
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
-      rescue Errno::ENOENT => _
-        CommandResult.new('', '', 1)
+      def login_command
+        nil # none, open your shell
+      end
+
+      def uri
```
chris-rock: Why are we exposing this one?

lib/train/transports/mock.rb
```diff
@@ -103,24 +104,26 @@ def command_not_found(cmd)
...
+    private
+
+    def run_command_via_connection(cmd)
+      @commands[cmd] ||
+        @commands[Digest::SHA256.hexdigest cmd.to_s] ||
+        command_not_found(cmd)
+    end
+
+    def file_via_connection(path)
+      @files[path] ||= file_not_found(path)
```
chris-rock: Should we use the file cache and just fill the cache? So this will become a noop then

lib/train/transports/docker.rb
jquick: I think we need it for now as inspec calls it here: https://github.com/chef/inspec/blob/master/lib/inspec/rspec_json_formatter.rb#L561

lib/train/plugins/base_connection.rb
jquick: Yes, moved.

jquick:
This is updated with all changes.

lib/train/plugins/base_connection.rb
```diff
@@ -15,18 +15,38 @@ class Train::Plugins::Transport
...
+      @cache_enabled.each_key do |type|
+        clear_cache(type)
+      end
+    end
+
+    def cache_enabled?(type)
+      @cache_enabled[type.to_sym]
+    end
+
+    def enable_cache(type)
```
chris-rock: I think we need a comment to document which types are available.

lib/train/plugins/transport.rb
```diff
@@ -1,7 +1,4 @@
 # encoding: utf-8
-#
-# Author:: Dominik Richter (<dominik.richter@gmail.com>)
```
chris-rock: I am not sure why this is removed?

jquick: It was a directive via Adam to remove any author tags on files we touch. I can add it back for now.

jquick:
This has been updated with changes. chef/inspec#2309 was also updated to use this new functionality.

lib/train/transports/docker.rb
chris-rock: Missed it, thank you.

lib/train/plugins/base_connection.rb
```diff
@@ -15,18 +15,40 @@ class Train::Plugins::Transport
...
+    end
+
+    def cache_enabled?(type)
+      @cache_enabled[type.to_sym]
+    end
+
+    # Enable caching types for Train. Currently we support
+    # :file and :command types
+    def enable_cache(type)
+      @cache_enabled[type.to_sym] = true
```
adamleff: Future enhancement: raise an exception if the user tries to enable/disable a cache that we don't support.

lib/train/plugins/base_connection.rb
```diff
@@ -52,39 +74,37 @@ def local?
...
-    # from files via the transport.
-    #
-    # @param [String] path which is being inspected
-    # @return [FileCommon] file object that allows for interaction
-    def file(_path, *_args)
-      fail Train::ClientError, "#{self.class} does not implement #file(...)"
+    # This is the main command call for all connections. This will call the private
+    # run_command_via_connection on the connection with optional caching
+    def run_command(cmd)
+      return @cache[:command][cmd] ||= run_command_via_connection(cmd) if cache_enabled?(:command)
```
adamleff: I think it might be a bit cleaner to read if it was structured to do the non-cache enabled stuff first:

```ruby
return run_command_via_connection(cmd) unless cache_enabled?(:command)

@cache[:command][cmd] ||= run_command_via_connection(cmd)
```

The `return` combined with the `||=` is hanging me up a bit. Feels like too much happening in one conditional statement.

lib/train/plugins/base_connection.rb
```diff
@@ -52,39 +74,37 @@ def local?
...
+
+      run_command_via_connection(cmd)
+    end
+
+    # This is the main file call for all connections. This will call the private
+    # file_via_connection on the connection with optional caching
+    def file(path, *args)
+      return @cache[:file][path] ||= file_via_connection(path) if cache_enabled?(:file)
```
adamleff: Same feedback as the `run_command` structure feedback.

lib/train/plugins/base_connection.rb
chris-rock: +1 great catch!

lib/train/plugins/base_connection.rb
chris-rock: Uh, I like that one. Thank you @adamleff

lib/train/plugins/base_connection.rb
```diff
@@ -43,10 +43,12 @@ def cache_enabled?(type)
     # Enable caching types for Train. Currently we support
     # :file and :command types
     def enable_cache(type)
+      fail Train::UnknownCacheType, "#{type} is not a valid cache type" unless @cache_enabled.keys.include?(type.to_sym)
```
adamleff: Rather than getting a list of the keys, you can ask the Hash if it has the key:

```ruby
... unless @cache_enabled.key?(type.to_sym)
```

PR #216, opened 2017-11-20
Fix inspec mock tests

jquick:
Inspec has some unique test cases where it sets nil platform name and uses a mac_os_x platform in some cases. I have created a actual mac_os_x platform and refined darwin checks. This change also updates mock os properties to be unknown if set to nil.

Inspec also expects a darwin family for its cases and documentation. That was added.

The new platform checks have been tested on darwin/osx boxes. Local Inspec tests are all passing with this change.

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -425,16 +425,31 @@ module Train::Platforms::Detect::Specifications
             # for now we are going to just try each platform
             true
           }
-      plat.name('darwin').title('Darwin').in_family('bsd')
+      plat.family('darwin').in_family('bsd')
           .detect {
             cmd = unix_file_contents('/usr/bin/sw_vers')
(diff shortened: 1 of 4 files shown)


lib/train/platforms/detect/specifications/os.rb
```diff
@@ -425,17 +425,32 @@ def self.load
...
+                @platform[:release] = m.nil? ? nil : m[1]
+                m = cmd.match(/^BuildVersion:\s+(.+)$/)
+                @platform[:build] = m.nil? ? nil : m[1]
+              end
+              @platform[:release] = unix_uname_r.lines[0].chomp if @platform[:release].nil?
               @platform[:arch] = unix_uname_m
               true
             end
           }
+      plat.name('mac_os_x').title('MAC OS X').in_family('darwin')
```
chris-rock: I think "Mac OS X is not the official spelling :-) I think it is `macOS X`

PR #220, opened 2017-11-21
Use named pipe to decrease local Windows runtime

jerryaldrichiii:
This uses PowerShell to spawn a process that listens on a named pipe for
Base64 encoded commands and executes them. This drastically decreases
the total runtime when running locally on Windows by using a single
PowerShell session to execute commands instead of spawning a new session
for each command.

Huge thanks to @sdelano for the initial idea!

Signed-off-by: Jerry Aldrich <jerryaldrichiii@gmail.com>

### DIFF
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -16,28 +16,18 @@ module Train::Transports
       @connection ||= Connection.new(@options)
     end
 
-    class Connection < BaseConnection # rubocop:disable Metrics/ClassLength
-      require 'json'
-      require 'base64'
-      require 'securerandom'
-
+    class Connection < BaseConnection
       def initialize(options)
         super(options)
         @cmd_wrapper = nil
         @cmd_wrapper = CommandWrapper.load(self, options)
-        if @platform.windows?
-          begin
-            @pipe = acquire_named_pipe
-          rescue
-            @pipe = false
-          end
-        end
+        @platform = platform
       end
 
       def run_command(cmd)
-        if defined?(@pipe) && @pipe
-          res = run_powershell_using_named_pipe(cmd)
-          CommandResult.new(res['stdout'], res['stderr'], res['exitstatus'])
+        if defined?(@platform) && @platform.windows?
+          @windows_runner ||= WindowsRunner.new
+          @windows_runner.run_command(cmd)
         else
           cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
           res = Mixlib::ShellOut.new(cmd)
@@ -53,12 +43,11 @@ module Train::Transports
       end
 
       def file(path)
-        @files[path] ||= \
-          if os.windows?
-            Train::File::Local::Windows.new(self, path)
-          else
-            Train::File::Local::Unix.new(self, path)
-          end
+        @files[path] ||= if os.windows?
+                           Train::File::Local::Windows.new(self, path)
+                         else
+                           Train::File::Local::Unix.new(self, path)
+                         end
       end
 
       def login_command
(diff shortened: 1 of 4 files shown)


lib/train/transports/local.rb
```diff
@@ -16,18 +16,27 @@ def connection(_ = nil)
...
         @cmd_wrapper = CommandWrapper.load(self, options)
       end
 
       def run_command(cmd)
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        if defined?(@platform) && @platform.windows?
+          start_named_pipe_server unless File.exist?('//localhost/pipe/InSpec')
```
chris-rock: As highlighted by @adamleff we should make the pipe unique

jerryaldrichiii:
Good catch @chris-rock @adamleff. I have fixed this to append a hex string to the end of the pipe name.

lib/train/transports/local.rb
```diff
@@ -36,6 +48,95 @@ def local?
...
       end
 
+      def run_powershell_using_named_pipe(pipe_location, script)
+        pipe = nil
+        # Try to acquire pipe for 10 seconds with 0.1 second intervals.
+        # Removing this can result in instability due to the pipe being
+        # temporarily unavailable.
+        100.times do
+          begin
+            pipe = open(pipe_location, 'r+')
```
chris-rock: Do we need to close the pipe?

lib/train/transports/local.rb
```diff
@@ -36,6 +48,95 @@ def local?
...
+        encoded_script = Base64.strict_encode64(script)
+        pipe.puts(encoded_script)
+        pipe.flush
+        result = JSON.parse(Base64.decode64(pipe.readline))
+        pipe.close
+        result
+      end
+
+      def start_named_pipe_server(pipe_name) # rubocop:disable Metrics/MethodLength
+        require 'win32/process'
```
chris-rock: Can we implement a fallback for windows platforms where this is not available, eg. windows nano?, in this case we have powershell available already.

lib/train/transports/local.rb
jerryaldrichiii: Interesting idea...after consulting with @jquick I think we should be able to do that. I'll get to work!

lib/train/transports/local.rb
jerryaldrichiii: Is there a way I can identify Nano with current Train? If not I can fallback to not creating a pipe if I can't open one. Thoughts?

lib/train/transports/local.rb
```diff
@@ -17,17 +17,48 @@ def connection(_ = nil)
...
+      require 'securerandom'
+
       def initialize(options)
         super(options)
         @cmd_wrapper = nil
         @cmd_wrapper = CommandWrapper.load(self, options)
+        @pipe = acquire_named_pipe if @platform.windows?
+      end
+
+      def acquire_named_pipe
```
chris-rock: I think that should be a private function

lib/train/transports/local.rb
```diff
@@ -17,17 +17,48 @@ def connection(_ = nil)
...
 
       def run_command(cmd)
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        if defined?(@pipe)
+          res = run_powershell_using_named_pipe(cmd)
+          CommandResult.new(res['stdout'], res['stderr'], res['exitstatus'])
+        else
```
chris-rock: I think this is problematic, we check if a pipe is available. Lets assume we are on a windows platform and pipe creation failed, then we need to implement a fallback. Essentially we need to decide:
```
if  os.windows?
  if pipe
    # use pipe 
  else
    # fallback, which may be slower but that is okay
  end
else 
  # linux / mac / unix
end 
```

lib/train/transports/local.rb
```diff
@@ -36,6 +67,60 @@ def local?
         true
       end
 
+      def run_powershell_using_named_pipe(script)
```
chris-rock: run and start pipe need to be private functions

lib/train/transports/local.rb
```diff
@@ -36,6 +67,60 @@ def local?
...
+        script = "$ProgressPreference='SilentlyContinue';" + script
+        encoded_script = Base64.strict_encode64(script)
+        @pipe.puts(encoded_script)
+        @pipe.flush
+        JSON.parse(Base64.decode64(@pipe.readline))
+      end
+
+      def start_named_pipe_server(pipe_name)
+        require 'win32/process'
```
chris-rock: Is there a reason why we are not using mixlib-shellout, since we just start a powershell command?

jerryaldrichiii: I'm using `win32/process` to spawn a background process. Can I use `mixlib-shellout` to start a background process?

Since the PowerShell server script never returns (essentially a daemon) I'm not sure I can use `mixlib-shellout`.

lib/train/transports/local.rb
jerryaldrichiii: I'll bring back the old functionality and then wrap `acquire_named_pipe` in a `begin/rescue`. That way if the pipe fails for whatever reason, it will default to the way things run now.

lib/train/transports/local.rb
```diff
@@ -21,13 +21,19 @@ def initialize(options)
...
+        @platform = platform
       end
 
       def run_command(cmd)
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        if defined?(@platform) && @platform.windows?
+          @windows_runner ||= WindowsRunner.new
```
chris-rock: The `WindowsRunner` is a great step in the right direction. Should we do this analog to our file handling? eg. 
```
Train::Command::Local::UnixRunner.new()
Train::Command::Local::WindowsPipeRunner.new()
# Fallback when the pipe is not working
Train::Command::Local::WindowsShellRunner.new()
```

lib/train/transports/local.rb
```diff
@@ -52,6 +57,133 @@ def login_command
...
+        require 'json'
+        require 'base64'
+        require 'securerandom'
+
+        def initialize
+          @pipe = acquire_pipe
+        end
+
+        def run_command(cmd)
+          res = @pipe ? run_via_pipe(cmd) : run_via_shellout(cmd)
```
chris-rock: I would split the runner into two, so that we have a `WindowsPipeRunner` and a `WindowsShellRunner`. This would make the implementation more clean. It also allows us later to run the pipe only on supported windows systems.

lib/train/transports/local.rb
```diff
@@ -52,6 +57,133 @@ def login_command
...
         'local://'
       end
+
+      class WindowsRunner
+        require 'json'
+        require 'base64'
+        require 'securerandom'
+
+        def initialize
+          @pipe = acquire_pipe
```
chris-rock: I would just throw a train exception that can be handled and we have a proper fallback, which would be the `WindowsShellRunner`

lib/train/transports/local.rb
jerryaldrichiii: I like the idea of separating the Windows runners. I don't see a reason for a `UnixRunner` at this time. I may create a `GenericRunner` which will use `ShellOut` and be used by Linux and the first few commands that determine the OS.

chris-rock: The unixrunner was the default shell out :-)

lib/train/transports/local.rb
jerryaldrichiii: Agreed! I make it so in the latest commits.

lib/train/transports/local.rb
jerryaldrichiii: Yup, but a few commands are ran using ShellOut while the OS is being identified, that's why I opted for `GenericRunner`.

lib/train/transports/local.rb
```diff
@@ -21,13 +23,18 @@ def initialize(options)
...
       def run_command(cmd)
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        if defined?(@os) && @os.windows?
+          @windows_runner ||= WindowsRunner.new
+          @windows_runner.run_command(cmd)
+        else
+          cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
```
chris-rock: Should we move the command wrapper handling into the GenericRunner now?

lib/train/transports/local.rb
```diff
@@ -21,13 +23,18 @@ def initialize(options)
...
+        @os = os
       end
 
       def run_command(cmd)
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
-        res = Mixlib::ShellOut.new(cmd)
-        res.run_command
-        CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        if defined?(@os) && @os.windows?
+          @windows_runner ||= WindowsRunner.new
```
chris-rock: should we decide for the runner during initialization? we could just can `@runner.run_command(cmd)` then here

lib/train/transports/local.rb
```diff
@@ -52,6 +59,154 @@ def login_command
...
+
+      class GenericRunner
+        def run_command(cmd)
+          res = Mixlib::ShellOut.new(cmd)
+          res.run_command
+          Local::CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        end
+      end
+
+      class WindowsRunner
```
chris-rock: if we move the creation to the initialize phase of `Connection`, I think we do not need the `WindowsRunner` abstraction.

lib/train/transports/local.rb
chris-rock: So much cleaner now, thank you!

lib/train/transports/local.rb
```diff
@@ -53,6 +63,147 @@ def file_via_connection(path)
...
+          @pipe.puts(encoded_script)
+          @pipe.flush
+          res = OpenStruct.new(JSON.parse(Base64.decode64(@pipe.readline)))
+          Local::CommandResult.new(res.stdout, res.stderr, res.exitstatus)
+        end
+
+        private
+
+        def acquire_pipe
+          current_pipe = Dir.entries('//./pipe/').find { |f| f =~ /inspec_/ }
```
adamleff: I'm a little concerned about this logic. What if there's another InSpec running on the host at the same time this one attempts to acquire a pipe? What if there's two? We could potentially find a pipe that's not ours and start using it only to have it disappear between the time we've acquired it and the time we go to use it.

Maybe my lack of knowledge around windows and named pipes is misleading my judgment here... but wouldn't it be better if we stored the name of the pipe we intend to use, look for it when we want to use it, and then if it's not there or not usable for some reason, create it? Some half-baked code to illustrate my idea...

```ruby
def pipe_name
  @pipe_name ||= "inspec_#{SecureRandom.hex}"
end

def reset_pipe_name
  @pipe_name = nil
end

def acquire_pipe
  pipe = open("//./pipe/#{pipe_name}", ...)
rescue
  create_pipe
end

def create_pipe
  # don't try to create a pipe of the same name if it exists but we can't open it
  reset_pipe_name if File.exist?(pipe_name)

  # do the creation and stuff, return nil if we can't create
end
```

jerryaldrichiii: I don't like the logic either.

Good news, while trying to implement your solution I realized something. We don't need to open an existing pipe! The `acquire_pipe` method only gets called once per object so there is no need to attempt to open an existing pipe. The pipe shouldn't be shared between separate processes in any scenario.

I'll get that fixed.

test/windows/local_test.rb
```diff
@@ -40,6 +40,33 @@
     cmd.stderr.must_equal ''
   end
 
+  it 'uses a named pipe if available' do
```
adamleff: I don't see anything obvious in this test that actually tests that the named pipe is used. How are you asserting that in this test?

test/windows/local_test.rb
```diff
@@ -40,6 +40,33 @@
...
+    # Must call `:conn` early so we can stub `SecureRandom`
+    connection = conn
+
+    SecureRandom.stubs(:hex).returns('with_pipe')
+    cmd = connection.run_command('Get-ChildItem //./pipe/ | Where-Object { $_.Name -Match "inspec_with_pipe" }')
+    cmd.stdout.wont_be_nil
+    cmd.stderr.must_equal ''
+  end
+
+  it 'when named pipe is not available it runs `Mixlib::Shellout`' do
```
adamleff: I don't see anything obvious in this test that actually tests that Mixlib::ShellOut is used instead of the named pipe. How are you asserting that in this test?

test/windows/local_test.rb
jerryaldrichiii: Great point, I added words to the comments to make it more clear. Basically, it runs a PowerShell command and checks if the pipe was created by looking for the file that would be created.

test/windows/local_test.rb
jerryaldrichiii: Great point, I added words to the comments to make it more clear. Basically, it runs a PowerShell command and that command checks to see if a pipe exists. If it does then we know it used the pipe and not `Mixlib::ShellOut`.

test/unit/transports/local_test.rb
```diff
@@ -93,4 +94,36 @@ def mock_run_cmd(cmd, &block)
       end
     end
   end
+
+  describe 'when running on Windows' do
+    let(:connection) { TransportHelper.new(:windows).transport.connection }
+
```
adamleff: Remove extra newline - it's more common to group all the `let`s together than space them out.

test/unit/transports/local_test.rb
```diff
@@ -6,10 +6,11 @@
 class TransportHelper
   attr_accessor :transport
 
-  def initialize
+  def initialize(type = :mock)
     Train::Platforms::Detect::Specifications::OS.load
-    plat = Train::Platforms.name('mock').in_family('linux')
+    plat = Train::Platforms.name('mock')
     plat.add_platform_methods
+    plat.stubs(:windows?).returns(true) if type == :windows
```
adamleff: The seems a bit brute-force. Can we accept an options hash and then we can plumb the family in as necessary and let the `Train::Platforms` stuff work normally without stubbing methods on this?

```ruby
def initialize(opts = {})
  Train::Platforms::Detect::Specifications::OS.load
  plat = Train::Platforms.name('mock')
  plat.in_family(opts[:family]) unless opts[:family].nil?
  plat.add_platform_methods
end
```

Would that work?

jerryaldrichiii: Sure will! Great idea. See latest commit.

test/unit/transports/local_test.rb
```diff
@@ -6,11 +6,12 @@
 class TransportHelper
   attr_accessor :transport
 
-  def initialize(type = :mock)
+  def initialize(user_opts = {})
+    opts = {platform_name: 'mock', family_hierarchy: ['mock']}.merge(user_opts)
     Train::Platforms::Detect::Specifications::OS.load
-    plat = Train::Platforms.name('mock')
+    plat = Train::Platforms.name(opts[:platform_name])
+    plat.family_hierarchy = opts[:family_hierarchy]
```
adamleff: Does `plat.family_hierarchy` need to be an array? The default you have set makes me believe that maybe that's true...

If so, can I suggest changing this to:

```ruby
plat.family_hierarchy = Array(opts[:family_hierarchy])
```

... so that it's always an array, even if a future test author supplies a single family string?

jerryaldrichiii: Based on a `grep` I did I see it defined as an array, but on the `mock` transport I see we do a flatten.

https://github.com/chef/train/blob/master/lib/train/transports/mock.rb#L80

I think I'll do what you said, but leave my examples as single item arrays.

Thoughts?

PR #225, opened 2017-12-06
Allow runner specifications for local connections

jerryaldrichiii:
This allows for overriding the default runner selection. For example, when running InSpec unit tests via AppVeyor on Windows the mock file loads would create too many named pipe processes thus exhausting available resources.

### DIFF
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -20,8 +20,8 @@ module Train::Transports
       def initialize(options)
         super(options)
 
-        @runner = if options[:runner]
-                    force_runner(options[:runner])
+        @runner = if options[:command_runner]
+                    force_runner(options[:command_runner])
                   else
                     select_runner(options)
                   end
@@ -54,16 +54,16 @@ module Train::Transports
         end
       end
 
-      def force_runner(runner)
-        case runner
-        when 'generic'
+      def force_runner(command_runner)
+        case command_runner
+        when :generic
           GenericRunner.new(self, options)
-        when 'windows_pipe'
+        when :windows_pipe
           WindowsPipeRunner.new
-        when 'windows_shell'
+        when :windows_shell
           WindowsShellRunner.new
         else
-          fail "Runner type \'#{runner}\' not supported"
+          fail "Runner type `#{command_runner}` not supported"
         end
       end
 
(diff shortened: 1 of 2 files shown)


lib/train/transports/local.rb
```diff
@@ -50,8 +41,39 @@ def uri
...
+
+      def force_runner(runner)
+        case runner
+        when 'generic'
+          GenericRunner.new(self, options)
+        when 'windows_pipe'
+          WindowsPipeRunner.new
+        when 'windows_shell'
+          WindowsShellRunner.new
+        end
```
adamleff: We should probably raise an exception if a runner was supplied that we don't support.

lib/train/transports/local.rb
```diff
@@ -50,8 +41,39 @@ def uri
...
+      end
+
       def run_command_via_connection(cmd)
-        @runner.run_command(cmd)
+        if defined?(@runner)
```
adamleff: If we add the exception handling above, do we need this change? Would there ever be a case where we didn't have a valid `@runner`?

jerryaldrichiii: Yes, the first few commands run do not use `GenericRunner`, but instead just use `ShellOut`.

The difference being that GenericRunner attempts to do the `CommandWrapper` bits.

lib/train/transports/local.rb
```diff
@@ -22,18 +20,11 @@ class Connection < BaseConnection
...
-
-        if os.windows?
-          # Attempt to use a named pipe but fallback to ShellOut if that fails
-          begin
-            @runner = WindowsPipeRunner.new
-          rescue PipeError
-            @runner = WindowsShellRunner.new
-          end
-        end
+        @runner = if options[:runner]
```
adamleff: I also wonder if this should be `:command_runner` and not just `:runner` in case we have file runners, etc. in the future.

lib/train/transports/local.rb
jerryaldrichiii: Good catch, will add an exception.

lib/train/transports/local.rb
jerryaldrichiii: Love it. Much nicer! Will make it so.

lib/train/transports/local.rb
jerryaldrichiii: Don't see any reason not to do it. Will make it so.

PR #227, opened 2017-12-10
ProxyCommand support

cbeckr:
This adds basic support for connecting via jump hosts.

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -113,7 +113,7 @@ class Train::Transports::SSH
         message: "Waiting for SSH service on #{@hostname}:#{@port}, " \
                  "retrying in #{delay} seconds",
       )
-      execute(PING_COMMAND.dup)
+      run_command(PING_COMMAND.dup)
     end
 
     def uri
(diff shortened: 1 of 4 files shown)


jquick:
@cbeckr - Thank you so much for your first contribution! Welcome to the family :)

This looks great and I will test it out soon and get back to you!

lib/train/transports/ssh_connection.rb
```diff
@@ -113,7 +114,7 @@ def wait_until_ready
         message: "Waiting for SSH service on #{@hostname}:#{@port}, " \
                  "retrying in #{delay} seconds",
       )
-      execute(PING_COMMAND.dup)
+      run_command(PING_COMMAND.dup)
```
jquick: I think we want to use the `run_command_via_connection` here instead. The run_command has the potential for caching which could cause unexpected results.

lib/train/transports/winrm_connection.rb
```diff
@@ -80,7 +80,7 @@ def wait_until_ready
         retry_limit: @max_wait_until_ready / delay,
         retry_delay: delay,
       )
-      execute(PING_COMMAND.dup)
+      run_command(PING_COMMAND.dup)
```
jquick: Same as above, I think it would be safer to `run_command_via_connection`

jquick:
I think this looks good for now, I plan on updating all the "wait_until_ready" flags as its like this on other transports also.

PR #228, opened 2017-12-14
Return platform name that is lower case and underscored

### LINKED ISSUE #191: platform names should be lower case
Running detect on some operating systems returns upper case values:
```
inspec detect -t ssh://vagrant@127.0.0.1:2200 -i /Users/chartmann/Development/tests/freebsd-11/.vagrant/machines/default/virtualbox/private_key 

== Operating System Details

Name:      FreeBSD
Family:    freebsd
Release:   11.1-RELEASE-p1
Arch:      
```
This leads to wrong behavior where we do checks like
```
%w{freebsd}.include?(os[:name])
```

jquick:
Fixes #191
Fixes chef/inspec#1732

This change is a remake of https://github.com/chef/train/pull/192 with the updated platform work. This change sets platform names to always return downcase and underscored.

### DIFF
--- a/lib/train/platforms/platform.rb
+++ b/lib/train/platforms/platform.rb
@@ -3,7 +3,7 @@
 module Train::Platforms
   class Platform
     include Train::Platforms::Common
-    attr_accessor :backend, :condition, :families, :family_hierarchy, :name_updated, :platform
+    attr_accessor :backend, :condition, :families, :family_hierarchy, :platform
 
     def initialize(name, condition = {})
       @name = name
@@ -13,7 +13,6 @@ module Train::Platforms
       @platform = {}
       @detect = nil
       @title = name.to_s.capitalize
-      clean_name
 
       # add itself to the platform list
       Train::Platforms.list[name] = self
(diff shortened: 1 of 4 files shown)


lib/train/platforms/platform.rb
```diff
@@ -25,7 +26,18 @@ def direct_families
     def name
       # Override here incase a updated name was set
       # during the detect logic
-      @platform[:name] || @name
+      @clean_name
+    end
+
+    def clean_name
+      name = (@platform[:name] || @name)
+      # This is a history of name change being used upstream in inspec
```
adamleff: I don't know what this comment means.

lib/train/platforms/platform.rb
```diff
@@ -25,7 +26,18 @@ def direct_families
...
+      if name =~ /[A-Z ]/
+        @name_updated = [name]
+        name = name.downcase.tr(' ', '_')
+        @name_updated << name
```
adamleff: What is `@name_updated` used for?

lib/train/platforms/platform.rb
```diff
@@ -25,7 +26,18 @@ def direct_families
     def name
       # Override here incase a updated name was set
       # during the detect logic
-      @platform[:name] || @name
+      @clean_name
```
adamleff: It's a bit confusing to me to have an instance_var named `@clean_name` and a method named `clean_name` and then use them in different ways.

Can we just default to using the `clean_name` method, and memoize the cleaned_name in that method so we only do the translation once?

jquick:
⚠️  DO NOT MERGE ⚠️ 
This has to go out with a corresponding inspec PR which is still to come.

chris-rock:
This will break Windows checks, right?

jquick:
The inspec PR will keep it from breaking and just warn for now till inspec 2.0

adamleff:
@chris-rock this is totally a breaking change for Train, but SemVer-wise we're OK since this is still `0.x`.

We've got an InSpec change in the pipeline that will gracefully handle it there, giving warnings for users as we initially discussed a month or so ago.

We'll want to cut a Train release with this change, and then cut an InSpec release that pins on that with the deprecation handling @jquick is working on.

PR #231, opened 2018-01-02
Add the base platform to the platform hierarchy

jquick:
This is a small change that will add the base platform name to the family hierarchy. This change will be used upstream to allow InSpec "in_family?" call to be used with the platform name as well.

### DIFF
--- a/test/unit/platforms/detect/scanner_test.rb
+++ b/test/unit/platforms/detect/scanner_test.rb
@@ -12,7 +12,7 @@ describe 'scanner' do
     it 'return child' do
       family = Train::Platforms.family('linux')
       scanner.scan_family_children(family).name.must_equal('linux')
-      scanner.instance_variable_get(:@family_hierarchy).must_equal(['linux'])
+      scanner.instance_variable_get(:@family_hierarchy).must_equal(['linux', 'linux'])
     end
 
     it 'return nil' do
--- a/lib/train/platforms/detect/scanner.rb
+++ b/lib/train/platforms/detect/scanner.rb
@@ -46,7 +46,10 @@ module Train::Platforms::Detect
 
         if plat.class == Train::Platforms::Platform
           @platform[:family] = parent.name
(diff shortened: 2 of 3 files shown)


test/unit/platforms/detect/scanner_test.rb
```diff
@@ -12,7 +12,7 @@
     it 'return child' do
       family = Train::Platforms.family('linux')
       scanner.scan_family_children(family).name.must_equal('linux')
-      scanner.instance_variable_get(:@family_hierarchy).must_equal(['linux'])
+      scanner.instance_variable_get(:@family_hierarchy).must_equal(['linux', 'linux'])
```
adamleff: This seems.... weird. Should we be de-duping this? In what specific use cases would we want the family hierarchy to be doubled-up like this?

jquick: Yeah this could be a little odd when the platform name and the family name are the same. Some use cases which may never come up are:
1) If we ever are taking the count of this variable to see how deep into the specifications "stack" the platform lives.
2) If we are expecting a full map of the platform in this variable. Platform + all families.

This variable will likely never be used raw by the end user as we have helper methods which look at the contents. I personal like it "as is" as I prefer the exact hierarchy listings so I know exactly where this platform lives.

We could also say that the platform does not belong in the "family" hierarchy. We could just override the in_family? in inspec to also allow the platform name to be valid. Which may be cleaner.

adamleff: Yeah, I think your latter instinct is accurate. The platform name really isn't a "family" but this treats it as such, which feels like we're muddying the concepts.

I think the name and the family should remain distinct, and if consumers of this data (such as InSpec) need more flexible interpretation of that data, then they should implement it on their side.

jquick: Sounds good, Ill close this out and we can do it downstream.

PR #236, opened 2018-01-24
Update Train::File to perform md5sum and sha256sum through train.connection

### LINKED ISSUE #235: sha256sum and md5sum to execute on the remote machine
A discussion in the Slack #inspec channel leads to the idea that md5sum and sha256sum should be done on the remote machine to minimize encoding issues as well as long executions (file is transferred on unknown connection speeds).

md5sum: https://github.com/chef/train/blob/master/lib/train/file.rb#L57
sha256sum: https://github.com/chef/train/blob/master/lib/train/file.rb#L65

Discussion:
Slack: https://chefcommunity.slack.com/archives/C1XB6U6MN/p1516743917000525
Github: https://github.com/chef/inspec/issues/1898#issuecomment-306853524

tarcinil:
fixes #235

### DIFF
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -11,8 +11,6 @@ require 'train/file/remote/unix'
 require 'train/file/remote/linux'
 require 'train/file/remote/windows'
 require 'train/file/remote/qnx'
-require 'digest/sha2'
-require 'digest/md5'
 require 'train/extras/stat'
 
 module Train
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -55,17 +55,21 @@ module Train
     end
 
     def md5sum
-      res = Digest::MD5.new
-      res.update(content)
-      res.hexdigest
+      value = @backend.run_command("md5sum #{@path}").stdout.slice(0..31)
+      if value.empty?
+        value = nil
+      end
+      value
     rescue TypeError => _
       nil
     end
 
     def sha256sum
-      res = Digest::SHA256.new
-      res.update(content)
-      res.hexdigest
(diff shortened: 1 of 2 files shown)


tarcinil:
While the checks seem to pass, I would double-check.  Travis is showing issues.

https://travis-ci.org/chef/train/jobs/332602640#L750

tarcinil:
based on the test

```
  1) Failure:
file interface::block device#test_0010_has no md5sum [tests/path_block_device_test.rb:46]:
--- expected
+++ actual
@@ -1 +1 @@
-"d41d8cd98f00b204e9800998ecf8427e"
+nil
```
This should likely be updated to expect `nil` as the name of the test is ` interface::block device#test_0010_has no md5sum`

lib/train/file.rb
```diff
@@ -55,17 +53,21 @@ def type
...
     def md5sum
-      res = Digest::MD5.new
-      res.update(content)
-      res.hexdigest
+      value = @backend.run_command("md5sum #{@path}").stdout.slice(0..31)
+      if value.empty?
+        value = nil
+      end
+      value
     rescue TypeError => _
```
tarcinil: Are there any other classes that I should rescue from?  I was thinking of those in https://github.com/chef/train/blob/master/lib/train/errors.rb

chris-rock:
I fully agree with the idea to make move the calculation to the os without the need to load the file in memory. I agree with @adamleff we need to make a platform dependent choice for the local tooling to calculate the md5/sha sums.

tarcinil:
superseded by #251

PR #239, opened 2018-01-29
Add AWS transport

### LINKED ISSUE #229: Add aws:// target
🚧  DO NOT MERGE until chef/inspec-aws#149 is resolved 🚧 

The inspec-aws resource pack currently works "by accident"; all resources ignore the inspec runner (and anything it knows about the transport), and simply read credentials from the AWS_* environment variables.  Inspec thinks it is running in local mode the whole time.

We've discussed internally having an explicit aws:// target, with possibilities of using various parts of the URI for various credentials.

This issue may need to be split, for an RFC to determine the URI format, then an implementation issue.

Refs https://github.com/chef/inspec/issues/2233

clintoncwolfe:
Adds a new type of platform, the `api` type.  AWS is added as a transport.  Reads credentials information from the constructor, from environment variables, or (when used from Inspec) via a URL.

URL format:
```
aws://     # Read AWS_REGION and AWS_PROFILE from env vars
aws://us-east-2  # Read AWS_PROFILE from env var
aws://us-east-2/my-profile   # Explicit.
```

Train users may access AWS client objects via `train.connection.aws_client(Aws::EC2::Client)` (for example).  The connection objects are cached, but their API call results are not.

Closes #229 .

### DIFF
--- a/lib/train/transports/aws.rb
+++ b/lib/train/transports/aws.rb
@@ -1,0 +1,72 @@
+require 'train/plugins'
+require 'aws-sdk'
+
+module Train::Transports
+  class Aws < Train.plugin(1)
+    name 'aws'
+    option :region, required: true, default: ENV['AWS_REGION']
+    option :access_key_id, default: ENV['AWS_ACCESS_KEY_ID']
+    option :secret_access_key, default: ENV['AWS_SECRET_ACCESS_KEY']
+    option :session_token, default: ENV['AWS_SESSION_TOKEN']
+
+    # This can provide the access key id and secret access key
+    option :profile, default: ENV['AWS_PROFILE']
+
+    def connection(_ = nil)
+      @connection ||= Connection.new(@options)
+    end
+
+    class Connection < BaseConnection
+      def initialize(options)
+        # Override for any cli options
+        # aws://region/my-profile
+        options[:region] = options[:host] || options[:region]
+        if options[:path]
+          # string the leading / from path
+          options[:profile] = options[:path][1..-1]
+        end
+        super(options)
+
+        @cache_enabled[:api_call] = true
+        @cache[:api_call] = {}
+
+        connect
+      end
+
+      def platform
+        direct_platform('aws')
+      end
+
+      def aws_client(klass)
+        return klass.new unless cache_enabled?(:api_call)
+        @cache[:api_call][klass.to_s.to_sym] ||= klass.new
+      end
+
+      def aws_resource(klass, args)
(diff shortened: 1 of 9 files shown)


lib/train/transports/aws.rb
```diff
@@ -0,0 +1,73 @@
...
+        return klass.new unless cache_enabled?(:api_call)
+        @cache[:api_call][klass.to_s.to_sym] ||= klass.new
+      end
+
+      def aws_resource(klass, args)
+        klass.new(args)
+      end
+
+      def connect
+        if @options[:profile]
```
adamleff: What if they have an `.aws/credentials` file with a `[default]` profile set and don't pass that in as the `profile` name here? Will `::Aws::Credentials.new` do the right thing with a bunch of `nil` parameters?

jquick: Currently if you do not supply a profile it will not default to a default one. It will just look at the env's and if those are nil you will have invalid credentials and get the following error when trying to request something.
`require_credentials': unable to sign request without credentials set (Aws::Errors::MissingCredentialsError)`

We could make it to where we allow aws to do its thing unless env's or a profile is set. This would allow you to use the default creds. To do this we would just want to update line 60 to ~
```
        opts = { region: @options[:region] }
        opts[:credentials] = creds unless creds.access_key_id.nil?
```
This will keep the default credentials unless we are actually updating them.

adamleff: I could’ve sworn that if you set up a client object and don’t set any credentials object in the config, it just does The Right Thing with regards to reading env vars, or reading the credentials file and using the “default” group by default.

test/unit/transports/aws_test.rb
```diff
@@ -90,5 +90,12 @@ def initialize(hash)
       creds[:credentials].session_token.must_equal 'test_session_token'
       creds[:region].must_equal 'test_region'
     end
+
+    it 'validate aws connection fallback' do
+      options[:profile] = nil
+      options[:access_key_id] = nil
+
+      Aws.config.expects(:update).never
```
adamleff: I don't think this test is doing anything. Don't you want to call `connection.connect`? And doesn't `Aws.config` always receive `:update` but in this case, it will be with options that don't include a `creds` key?

lib/train/transports/aws.rb
adamleff: Thanks - I think that's the right call so that people have a consistent UX with all things that use the AWS SDK, including Train.

PR #244, opened 2018-01-30
fix aix class

arielrolfo:
Creating this PR based on chat with Jerry Aldrich


adamleff:
This actually looks like a larger problem. The file `qnx.rb` is defining a class for AIX systems. So that's really the bug we need to solve.

I'll look into this deeper later today. I suspect this is an error introduced during the refactor of all of Train's "file" classes, and this class name just got incorrectly inputted into this file. I'll take care of submitting a new PR later that correctly addresses this.

adamleff:
@arielrolfo I have submitted PR #249 that should fix this. Sorry for the issue you encountered, and thanks for taking an initial stab at fixing it!

PR #246, opened 2018-01-31
cisco ios12 detection

arlimus:


lib/train/platforms/detect/specifications/os.rb
```diff
@@ -27,12 +27,27 @@ def self.load
...
+            if unix_uname_s =~ /./ && !unix_uname_s.start_with?('Line has invalid autocommand ')
               @platform[:arch] = unix_uname_m
               true
             end
           }
 
+      # cisco_ios family
+      plat.family('cisco').title('Cisco Family')
+          .detect {
+            true
```
adamleff: I wonder if we shouldn't change this to be something like `!cisco_show_version.nil?` so we don't attempt to do Cisco detection when not necessary.

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -27,12 +27,27 @@ def self.load
...
+          }
+
+      plat.name('cisco_ios').title('Cisco IOS').in_family('cisco')
+          .detect {
+            return unless v == cisco_show_version
```
adamleff: I think you want this to be a single `=`, right? in-line assignment? Otherwise, I don't know where `v` is getting set and then it will either bomb here or in the next line when you try to fetch from it.

lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -52,5 +52,11 @@ def unix_uname_m
       return @uname[:m] if @uname.key?(:m)
       @uname[:m] = command_output('uname -m')
     end
+
+    def cisco_show_version
+      res = command_output('show version')
```
adamleff: Can we memoify this so we can call `cisco_show_version` multiple times in a run without running the command multiple times?

Perhaps something like this:

```ruby
return @cisco_show_version if defined?(@cisco_show_version)

res = command_output('show version')
m = res.match(/^Cisco IOS Software, [^,]+? \(([^,]+?)\), Version (\d+\.\d+)/)
@cisco_show_version = m.nil? ? nil : { version: m[2], model: m[1], raw: res }
```

arlimus: i got a follow-up that does that in a more general way; if we can get this merged I'll follow i tup

lib/train/platforms/detect/specifications/os.rb
arlimus: let me do that in the follow-up too

lib/train/platforms/detect/specifications/os.rb
arlimus: ah darn linter caught me on that one, 1sec

arlimus:
superseded

PR #250, opened 2018-02-02
Add azure transport

jquick:
Fixes https://github.com/chef/train/issues/233

Adds the azure transport and credential parsing. Heavily based off https://github.com/chef/inspec-azure/blob/master/libraries/azure_backend.rb

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,7 +19,7 @@ group :test do
   gem 'rake', '~> 10'
   gem 'rubocop', '~> 0.36.0'
   gem 'simplecov', '~> 0.10'
-  gem 'concurrent-ruby', '~> 0.9'
+  gem 'concurrent-ruby', '~> 1.0'
 end
 
 group :integration do
--- a/lib/train/transports/azure.rb
+++ b/lib/train/transports/azure.rb
@@ -116,7 +116,7 @@ module Train::Transports
       private
 
       def parse_credentials_file # rubocop:disable Metrics/AbcSize
-        # If an INSPEC_AZURE_CREDS environment has been specified set the
+        # If an AZURE_CRED_FILE environment variable has been specified set the
         # the credentials file to that, otherwise set the one in home
         azure_creds_file = @options[:credentials_file]
         azure_creds_file = File.join(Dir.home, '.azure', 'credentials') if azure_creds_file.nil?
--- a/lib/train/transports/azure.rb
+++ b/lib/train/transports/azure.rb
@@ -76,7 +76,7 @@ module Train::Transports
         return unless File.readable?(azure_creds_file)
 
         credentials = IniFile.load(File.expand_path(azure_creds_file))
-        if !@options[:subscription_id].nil?
+        if @options[:subscription_id]
           id = @options[:subscription_id]
         elsif !ENV['AZURE_SUBSCRIPTION_NUMBER'].nil?
           subscription_number = ENV['AZURE_SUBSCRIPTION_NUMBER'].to_i
--- a/lib/train/transports/azure.rb
+++ b/lib/train/transports/azure.rb
@@ -1,0 +1,108 @@
+# encoding: utf-8
+
+require 'train/plugins'
+require 'ms_rest_azure'
+require 'azure_mgmt_resources'
+require 'inifile'
+
+module Train::Transports
+  class Azure < Train.plugin(1)
+    name 'azure'
+    option :tenant_id, default: ENV['AZURE_TENANT_ID']
+    option :client_id, default: ENV['AZURE_CLIENT_ID']
+    option :client_secret, default: ENV['AZURE_CLIENT_SECRET']
+    option :subscription_id, default: ENV['AZURE_SUBSCRIPTION_ID']
+
+    # This can provide the client id and secret
+    option :credentials_file, required: true, default: ENV['AZURE_CRED_FILE']
+
+    def connection(_ = nil)
+      @connection ||= Connection.new(@options)
+    end
+
+    class Connection < BaseConnection
+      def initialize(options)
+        # Override for any cli options
+        # azure://subscription_id
+        options[:subscription_id] = options[:host] || options[:subscription_id]
+        super(options)
+
+        @cache_enabled[:api_call] = true
+        @cache[:api_call] = {}
+
+        if @options[:client_secret].nil? && @options[:client_id].nil?
+          parse_credentials_file
+        end
+        connect
+      end
+
+      def platform
+        direct_platform('azure')
+      end
+
+      def azure_client(klass = nil)
+        # return the managment client by default
+        klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client if klass.nil?
+        return klass.new(@credentials) unless cache_enabled?(:api_call)
+
+        @cache[:api_call][klass.to_s.to_sym] ||= klass.new(@credentials)
+      end
+
+      def connect
+        provider = ::MsRestAzure::ApplicationTokenProvider.new(
+          @options[:tenant_id],
+          @options[:client_id],
+          @options[:client_secret],
+        )
+
+        @credentials = {
+          credentials: ::MsRest::TokenCredentials.new(provider),
+          subscription_id: @options[:subscription_id],
+          tenant_id: @options[:tenant_id],
+          client_id: @options[:client_id],
+          client_secret: @options[:client_secret],
+        }
+      end
+
+      def uri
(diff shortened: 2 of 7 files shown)


lib/train/transports/azure.rb
```diff
@@ -0,0 +1,108 @@
...
+module Train::Transports
+  class Azure < Train.plugin(1)
+    name 'azure'
+    option :tenant_id, default: ENV['AZURE_TENANT_ID']
+    option :client_id, default: ENV['AZURE_CLIENT_ID']
+    option :client_secret, default: ENV['AZURE_CLIENT_SECRET']
+    option :subscription_id, default: ENV['AZURE_SUBSCRIPTION_ID']
+
+    # This can provide the client id and secret
+    option :credentials_file, required: true, default: ENV['AZURE_CRED_FILE']
```
adamleff: Seems weird to mark this required but also provide a default. Is this actually a required parameter? Looks like we do some magic later in a separate method when parsing the creds file to assume a reasonable default file location.

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,108 @@
...
+        connect
+      end
+
+      def platform
+        direct_platform('azure')
+      end
+
+      def azure_client(klass = nil)
+        # return the managment client by default
+        klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client if klass.nil?
```
adamleff: Should we just put this in the method header instead?

```ruby
def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client)
```

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,108 @@
...
+      end
+
+      private
+
+      def parse_credentials_file # rubocop:disable Metrics/AbcSize
+        # If an INSPEC_AZURE_CREDS environment has been specified set the
+        # the credentials file to that, otherwise set the one in home
+        azure_creds_file = @options[:credentials_file]
+        azure_creds_file = File.join(Dir.home, '.azure', 'credentials') if azure_creds_file.nil?
+        return unless File.file?(azure_creds_file)
```
adamleff: Using `File.readable?` might be better here to avoid permissions errors, etc.

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,108 @@
...
+          if subscription_number > credentials.sections.length
+            raise format(
+              'Your credentials file only contains %s subscriptions.  You specified number %s.',
+              @credentials.sections.length,
+              subscription_number,
+            )
+          end
+          id = credentials.sections[subscription_number - 1]
+        else
+          raise 'Multiple credentials detected, please set a subscription id to use.' if credentials.sections.count > 1
```
adamleff: I'd give them a clue on how to do that...

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,108 @@
...
+          id = credentials.sections[0]
+        end
+
+        raise format('No credentials found for subscription number %s', id) if credentials.sections.empty? || credentials[id].empty?
```
adamleff: regular string interpolation would work fine here too, right? We don't use `format` all that often...

```ruby
raise "No credentials found for subscription number #{id}"
```

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,106 @@
...
+        # the credentials file to that, otherwise set the one in home
+        azure_creds_file = @options[:credentials_file]
+        azure_creds_file = File.join(Dir.home, '.azure', 'credentials') if azure_creds_file.nil?
+        return unless File.readable?(azure_creds_file)
+
+        credentials = IniFile.load(File.expand_path(azure_creds_file))
+        if !@options[:subscription_id].nil?
+          id = @options[:subscription_id]
+        elsif !ENV['AZURE_SUBSCRIPTION_NUMBER'].nil?
+          subscription_number = ENV['AZURE_SUBSCRIPTION_NUMBER'].to_i
```
jerryaldrichiii: Thoughts on changing this to `ENV['AZURE_SUBSCRIPTION_INDEX']` or similar? On the first read through I got this confused with `ID`...though could just be me.

jquick: Yeah I was thinking the same thing. But we have that ENV from the old inspec-azure and I am not sure we want to break people (if anyone is actually using it). I actually am in favor of removing the whole index based credentials. Rather have them specify a specific subscription_id.

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,106 @@
...
+
+      def parse_credentials_file # rubocop:disable Metrics/AbcSize
+        # If an INSPEC_AZURE_CREDS environment has been specified set the
+        # the credentials file to that, otherwise set the one in home
+        azure_creds_file = @options[:credentials_file]
+        azure_creds_file = File.join(Dir.home, '.azure', 'credentials') if azure_creds_file.nil?
+        return unless File.readable?(azure_creds_file)
+
+        credentials = IniFile.load(File.expand_path(azure_creds_file))
+        if !@options[:subscription_id].nil?
```
jerryaldrichiii: could this be written as `@options[:subscription_id]`?

test/unit/transports/azure_test.rb
```diff
@@ -0,0 +1,146 @@
...
+
+  describe 'options' do
+    it 'defaults to env options' do
+      options[:tenant_id].must_equal 'test_tenant_id'
+      options[:client_id].must_equal 'test_client_id'
+      options[:client_secret].must_equal 'test_client_secret'
+      options[:subscription_id].must_equal 'test_subscription_id'
+    end
+
+    it 'test options override' do
```
jerryaldrichiii: Could we modify the `it` blocks to describe intent. Example:
```
it 'allows for options override`
```

test/unit/transports/azure_test.rb
```diff
@@ -0,0 +1,146 @@
...
+      credentials[:client_secret].must_equal 'test_client_secret'
+      credentials[:subscription_id].must_equal 'test_subscription_id'
+    end
+  end
+
+  describe 'parse_credentials_file' do
+    let(:cred_file) do
+      require 'tempfile'
+      file = Tempfile.new('cred_file')
+      info = <<-INFO
```
jerryaldrichiii: Not sure if leading whitespace matters in `.ini`, but you could use a squiggly heredoc. E.g. `<<~INFO`

jquick: Yeah I tried that, some of the appvoyer stuff for train was using 2.2 and was bombing. So reverted to the old system.

test/unit/transports/azure_test.rb
```diff
@@ -0,0 +1,146 @@
...
+
+        [my_subscription_id2]
+        client_id = "my_client_id2"
+        client_secret = "my_client_secret2"
+        tenant_id = "my_tenant_id2"
+      INFO
+      file.write(info)
+      file.close
+      file
+    end
```
jerryaldrichiii: Could we add a newline here?

Gemfile
```diff
@@ -19,7 +19,7 @@ group :test do
   gem 'rake', '~> 10'
   gem 'rubocop', '~> 0.36.0'
   gem 'simplecov', '~> 0.10'
-  gem 'concurrent-ruby', '~> 0.9'
+  gem 'concurrent-ruby', '~> 1.0'
```
jerryaldrichiii: Since this is a major version bump, will it affect anything down stream? ChefDK for example? Or is that not an issue here?

jquick: Not to worried as its only in the test bracket. But with the pining it should not be an issue.

lib/train/transports/azure.rb
```diff
@@ -0,0 +1,153 @@
...
+        end
+
+        # return the api version for the type
+        @apis[resource_type]
+      end
+
+      private
+
+      def parse_credentials_file # rubocop:disable Metrics/AbcSize
+        # If an INSPEC_AZURE_CREDS environment has been specified set the
```
chris-rock: where is the INSPEC_AZURE_CREDS env variable used/read? Is this going to be done in InSpec? If so, we should not documented here. Since train does not know anything about inspec.

jquick: This comment is referring to the `ENV['AZURE_CRED_FILE']` Ill update so it has the proper name.

chris-rock: Awesome, that makes sense!

PR #251, opened 2018-02-06
Modify checksum logic to use system binaries

### LINKED ISSUE #235: sha256sum and md5sum to execute on the remote machine

jerryaldrichiii:
This modifies `md5sum`/`sha256sum` to use binaries on the target system vs Ruby on the machine running InSpec.

This fixes #235 
This closes #236 

Many thanks to @tarcinil for the initial work on this! It is greatly appreciated!

### DIFF
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -5,8 +5,10 @@
 
 require 'train/file/local'
 require 'train/file/remote'
-require 'digest/sha2'
-require 'digest/md5'
+require 'train/file/remote/unix'
+require 'train/file/remote/linux'
+require 'train/file/remote/windows'
+require 'train/file/remote/qnx'
 require 'train/extras/stat'
 
 module Train
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -131,12 +131,35 @@ module Train
     end
 
     def md5sum
+      return perform_ruby_checksum(method) if defined?(@use_ruby_checksum)
+
+      @md5_command ||= case @backend.os.family
+                       when 'darwin'
+                         # `-r` reverses output to match `md5sum` behavior
+                         'md5 -r'
+                       when 'solaris'
+                         'digest -a md5'
+                       else
+                         'md5sum'
+                       end
+
       return perform_checksum(:md5) unless @backend.os.family == 'windows'
 
       perform_checksum_windows(:md5)
     end
 
     def sha256sum
+      return perform_ruby_checksum(method) if defined?(@use_ruby_checksum)
+
+      @sha256_command ||= case @backend.os.family
+                          when 'darwin', 'hpux', 'qnx'
+                            'shasum -a 256'
+                          when 'solaris'
+                            'digest -a sha256'
+                          else
+                            'sha256sum'
+                          end
+
       return perform_checksum(:sha256) unless @backend.os.family == 'windows'
 
       perform_checksum_windows(:sha256)
@@ -144,35 +167,12 @@ module Train
 
     private
 
-    def md5_command
-      case @backend.os.family
-      when 'darwin'
-        # `-r` reverses output to match `md5sum`
-        'md5 -r'
-      when 'solaris'
-        'digest -a md5'
-      else
-        'md5sum'
-      end
-    end
-
-    def sha256_command
-      case @backend.os.family
-      when 'darwin', 'hpux', 'qnx'
-        'shasum -a 256'
-      when 'solaris'
-        'digest -a sha256'
-      else
-        'sha256sum'
-      end
-    end
-
     def perform_checksum(method)
       case method
       when :md5
-        cmd = "#{md5_command} #{@path}"
+        cmd = "#{@md5_command} #{@path}"
       when :sha256
-        cmd = "#{sha256_command} #{@path}"
+        cmd = "#{@sha256_command} #{@path}"
       end
 
       res = @backend.run_command(cmd)
@@ -194,6 +194,9 @@ module Train
     # Digest to perform the checksum. This is less efficient than using remote
     # system binaries and can lead to incorrect results due to encoding.
     def perform_ruby_checksum(method)
+      # This is used to skip attempting other checksum methods. If this is set
+      # then we know all other methods have failed.
+      @use_ruby_checksum = true
       case method
       when :md5
         res = Digest::MD5.new
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -8,7 +8,7 @@ require 'train/file/remote'
 require 'train/extras/stat'
 
 module Train
-  class File
+  class File # rubocop:disable Metrics/ClassLength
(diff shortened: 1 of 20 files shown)


lib/train/file.rb
```diff
@@ -147,5 +129,71 @@ def mounted?
...
+    def md5_command
+      case @backend.os.family
+      when 'darwin'
+        # `-r` reverses output to match `md5sum`
+        'md5 -r'
+      when 'solaris'
+        'digest -a md5'
+      else
+        'md5sum'
+      end
```
chris-rock: I think we need to think about the case where those commands are not available. In those case I would like to fall back to the old methods

jerryaldrichiii: Sounds good, added in last commit. Let me know what you think @chris-rock.

chris-rock:
We are getting closer @jerryaldrichiii !!! I am going to request one other round for improvement. At this point we are detecting platforms and available commands again and again. Why not keep our decision for one connection. Therefore we have many implementations of the hash generation eg.:
- md5
- md5sum
- rubymd5
...
Once we call the function the first time, we detect the right implementation and stick for to that decision while the connection stays available. Does this makes any sense?

jerryaldrichiii:
Sure does @chris-rock. Let me know if the recent commits achieve what you're looking for.

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+                          end
+
+      return perform_checksum(:sha256) unless @backend.os.family == 'windows'
+
+      perform_checksum_windows(:sha256)
+    end
+
+    private
+
+    def perform_checksum(method)
```
chris-rock: lets call this `perform_checksum_cmd`

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+      res = @backend.run_command(cmd)
+      return res.stdout.split("\r\n")[1].tr(' ', '') if res.exit_status == 0
+
+      perform_ruby_checksum(method)
+    end
+
+    # This pulls the content of the file to the machine running Train and uses
+    # Digest to perform the checksum. This is less efficient than using remote
+    # system binaries and can lead to incorrect results due to encoding.
+    def perform_ruby_checksum(method)
```
chris-rock: should this be called `perform_checksum_ruby` to keep names aligned?

lib/train/file.rb
chris-rock: Technically you can call this method without setting the command before. I think we should either move the command detection into this method or assume it as an argument to minimize function side-effects.

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+      when :md5
+        cmd = "#{@md5_command} #{@path}"
+      when :sha256
+        cmd = "#{@sha256_command} #{@path}"
+      end
+
+      res = @backend.run_command(cmd)
+      return res.stdout.split(' ').first if res.exit_status == 0
+
+      perform_ruby_checksum(method)
```
chris-rock: I would return an error for that method instead of calling `perform_ruby_checksum` here

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+    end
+
+    def perform_checksum_windows(method)
+      cmd = "CertUtil -hashfile #{@path} #{method.to_s.upcase}"
+
+      res = @backend.run_command(cmd)
+      return res.stdout.split("\r\n")[1].tr(' ', '') if res.exit_status == 0
+
+      perform_ruby_checksum(method)
```
chris-rock: As above, I would not perform the ruby checksum here, but return an error

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+                         'md5sum'
+                       end
+
+      return perform_checksum(:md5) unless @backend.os.family == 'windows'
+
+      perform_checksum_windows(:md5)
+    end
+
+    def sha256sum
+      return perform_ruby_checksum(method) if defined?(@use_ruby_checksum)
```
chris-rock: I think method needs to be defined here

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
 
       !mounted.nil? && !mounted.stdout.nil? && !mounted.stdout.empty?
     end
+
+    def md5sum
+      return perform_ruby_checksum(method) if defined?(@use_ruby_checksum)
```
chris-rock: should this be `perform_checksum_ruby(:md5)`?

lib/train/file.rb
```diff
@@ -147,5 +129,85 @@ def mounted?
...
+                         'md5 -r'
+                       when 'solaris'
+                         'digest -a md5'
+                       else
+                         'md5sum'
+                       end
+
+      return perform_checksum(:md5) unless @backend.os.family == 'windows'
+
+      perform_checksum_windows(:md5)
```
chris-rock: I think this can be a bit more simple, something like:
```
method = :md5
if @hash_ruby_fallback
  hash, err = perform_checksum_windows(method) if os.family == 'windows'
  hash, err  = perform_checksum_cmd(method) if os.family != 'windows'
  if err == nil
    return hash
  end
end
@hash_ruby_fallback = true
return perform_checksum_ruby(method)
```

lib/train/file.rb
jerryaldrichiii: Yup, I think that is fine. I've done a search and replace.

jerryaldrichiii:
I took a different approach @chris-rock but the feedback was super valuable. Let me know if it is too divergent from what you had in mind.

jerryaldrichiii:
I fixed the merge conflicts via the GUI. No sure how I feel, but hey we squash any way.

lib/train/file.rb
```diff
@@ -147,5 +129,82 @@ def mounted?
...
+                   @sha256_command ||= case @backend.os.family
+                                       when 'darwin', 'hpux', 'qnx'
+                                         'shasum -a 256'
+                                       when 'solaris'
+                                         'digest -a sha256'
+                                       else
+                                         'sha256sum'
+                                       end
+
+                   perform_checksum_linux(@sha256_command)
```
chris-rock: should this be `perform_checksum_unix`?

jerryaldrichiii: Can't see why not. I've fixed it.

PR #254, opened 2018-02-08
Support for Brocade FOS-based SAN devices

marcelhuth:
Hi,

I made a few changes that makes train able to detect FOS (Fabric OS) of Brocade Fibre-Channel switches.

The whole thing has currently a problem (stated in the code as well): The OS of these devices is linux based and does respond to `uname` command correctly when logging in as root user. If you log in with root, train will detect _Unix/Linux/Generic Linux_ what is not what we want here. If you log in with a different (normal) user, that user will get rbash shell as CLI and `uname` command is not available what makes it possible to "fall through" the unix detection and detect a brocade_fos platform.

Normally you won't log in using root user to those devices (vendor says it's not supported), nevertheless you could do this and it would break the detection. 

One workaround would be to handle brocade detection before unix, but that would cause `version` command run on all systems (except windows). I guess this is not really an option. Unfortunately I have no idea how to get around this problem currently.

Thanks for your feedback!

### DIFF
--- a/lib/train/platforms/detect/helpers/os_common.rb
+++ b/lib/train/platforms/detect/helpers/os_common.rb
@@ -53,6 +53,18 @@ module Train::Platforms::Detect::Helpers
       @uname[:m] = command_output('uname -m')
     end
 
+    def brocade_version
+      return @cache[:brocade] if @cache.key?(:brocade)
+      res = command_output('version')
+
+      m = res.match(/^Fabric OS:\s+v(\S+)$/)
+      unless m.nil?
+        return @cache[:brocade] = { version: m[1], type: 'fos' }
+      end
+
+      @cache[:brocade] = nil
+    end
+
     def cisco_show_version
       return @cache[:cisco] if @cache.key?(:cisco)
       res = command_output('show version')
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -288,6 +288,13 @@ module Train::Platforms::Detect::Specifications
             end
           }
 
+      # brocade family detected here if device responds to 'uname' command,
+      # happens when logging in as root
+      plat.family('brocade').title('Brocade Family').in_family('linux')
+          .detect {
+            !brocade_version.nil?
+          }
+
       # genaric linux
       # this should always be last in the linux family list
       plat.name('linux').title('Genaric Linux').in_family('linux')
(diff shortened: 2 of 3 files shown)


chris-rock:
@marcelhuth This is awesome! If it is detected as Linux here https://github.com/chef/train/blob/master/lib/train/platforms/detect/specifications/os.rb#L329, you could add a check for brocade just before, therefore it will test brocade before jumping into generic linux.

marcelhuth:
@chris-rock Thanks for your feedback.
Actually it would not work this way unfortunately. Normally you won't work with root user on that devices. If you are not root (the standard use case) the CLI won't get you a stdout from `uname` and so the device is not a unix-family for train. At the end an exception will be thrown here https://github.com/chef/train/blob/a169efeca11404c084dd99802fb9391101ee5e40/lib/train/platforms/detect/scanner.rb#L41

So the code currently work for the standard use case (normal user to login) at the moment. It gets complicated when using root (and there are people working with root on the devices, even though it's not supported/recommended).

jquick:
Looks good to me. Thanks for your work on this @marcelhuth!

marcelhuth:
Hi @jquick ,
thanks for approving this. Please do not merge anything at the moment... I'm still thinking about how to fix the problem with different logins that lead to different behavior.

As @chris-rock stated the check could be moved just before "Generic Linux", but we only reach that code when using `root` user. So I'm curently thinking about having two checks for brocade_fos. The first check has to be done just after unix family (if logging in with any user different from `root` we reach this code and anything is fine) the second check has to be done before "Generic Linux" (this will be reached when logging in with `root`). When the second check find a brocade_fos we have to change the previously detected os.family from "unix" to "brocade", and I don't know if this is good practice or if this could even work... 

A second possibility is to throw an exception for "unsupported user in OS" (or something similar) when having the second check true. 

I would like to avoid the situation where somebody depends on the os.platform is "unix" and does anything that is supported there that breaks the brocade device when using `root` user.

Do you have any ideas on that?

Thanks,
Marcel

marcelhuth:
Hi @jquick and @chris-rock 
I tried to fix the issue with the root login. It should not break the default "Generic Linux" case just skip the detection of Linux if the devices responds to Brocade's `version` command and then go on with the next platform. What do you think?
Thanks,
Marcel

jquick:
This works but gets us into a position of having to support the downstream platforms by ignoring it. Instead I think you had the right idea of putting the bash version right above linux in its own family then redefine the family again at the bottom. I am submitting my own refactor for arista which has the same issue. Ill post here the section, when I post it.

jquick:
Hey @marcelhuth  - take a look at the arista example:
https://github.com/chef/train/blob/master/lib/train/platforms/detect/specifications/os.rb#L100 - defined here for the linux bash
https://github.com/chef/train/blob/master/lib/train/platforms/detect/specifications/os.rb#L478 - defined again for the other shell

marcelhuth:
Hi @jquick - saw your PR already ;) As this is now in master, I'll get my stuff changed to fit to the new base os family layout. I'll refactor the bash vs. "normal" shell login in the way you did with Arista as well. I'll try to get that ready the next days. Thanks for helping.

jquick:
Looks good to me! Thanks for your work on this @marcelhuth !

PR #258, opened 2018-02-14
Add api sdk versions as platform release

jquick:
This PR makes the release call functional for api platforms. This is needed for the upstream inspec detect which expects a release field.

### DIFF
--- a/lib/train/transports/aws.rb
+++ b/lib/train/transports/aws.rb
@@ -31,11 +31,14 @@ module Train::Transports
         @cache_enabled[:api_call] = true
         @cache[:api_call] = {}
 
+        # additional platform details
+        @platform_details = { release: '2.0' }
+
         connect
       end
 
       def platform
-        direct_platform('aws')
+        direct_platform('aws', @platform_details)
       end
 
       def aws_client(klass)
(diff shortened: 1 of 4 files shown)


lib/train/transports/azure.rb
```diff
@@ -35,11 +35,15 @@ def initialize(options)
         if @options[:client_secret].nil? && @options[:client_id].nil?
           parse_credentials_file
         end
+
+        # additional platform details
+        @platform_details = { release: '0.15' }
```
clintoncwolfe: We might need to clarify that this is the azure_mgmt_resource gem version, NOT the Azure API version.  The Azure API allows you to use different API versions for different services, and the gem supports that.  So you could talk to the VM service at version 2017-07-03, and the AD service at version 2018-02-05 (those are arbitrary, but they are named after dates)

lib/train/transports/aws.rb
```diff
@@ -31,11 +31,14 @@ def initialize(options)
         @cache_enabled[:api_call] = true
         @cache[:api_call] = {}
 
+        # additional platform details
+        @platform_details = { release: '2.0' }
```
clintoncwolfe: Where is 2.0 coming from?

PR #267, opened 2018-03-14
Override unf_ext gme for Azure resource

jquick:
This change loads in unf_ext with the solaris fix. This is a temporary bind until:

knu/ruby-unf_ext#39

Can be merged/released accordingly.

Credit to @scotthain


jquick:
We are going to do this at a different level.

PR #270, opened 2018-03-16
Add platform uuid information.

### LINKED ISSUE #264: RFC: Generate unique uuid for platforms
### Description

Generate a unique identifier for platforms that can be used to see if your hitting the same box over multiple runs. See if we already have a uuid on the box (via chef client or something else) else create one based on the criteria below.

### Unix
* `cat /etc/machine-id` || `/var/lib/dbus/machine-id` ||  `/var/db/dbus/machine-id` if valid
 * create uuids like:
```
      hash = Digest::SHA1.new
      hash.update(string)
      ary = hash.digest.unpack("NnnnnN")
      ary[2] = (ary[2] & 0x0FFF) | (5 << 12)
      ary[3] = (ary[3] & 0x3FFF) | 0x8000
      "%08x-%04x-%04x-%04x-%04x%08x" % ary
```
taken from https://github.com/rails/rails/blob/a9dc45459abcd9437085f4dd0aa3c9d0e64e062f/activesupport/lib/active_support/core_ext/digest/uuid.rb#L16

* We could write a uuid to the disk if none exist. `/etc/machine-uuid` or local if that is not possible

#### OSX
hardware uuid `system_profiler SPHardwareDataType | awk '/UUID/ { print $3; }'` 

### Windows
`wmic csproduct get UUID`
`HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Cryptography\MachineGuid` in the registry
* We could write a uuid to the disk if none exist. 

### AWS API
Use the account_id

### Azure
Use subscription_id, fallback to tenant_id

jquick:
This change adds a uuid option for platforms. This is either pulled from a unique identifier already on the box or by creating one. You also have the option of setting a command to run for a unique id during the platform detect block.

Fixes #264 


jquick:
I think I’m going to move the logic to the detect helpers.

lib/train/platforms/uuid.rb
```diff
@@ -0,0 +1,135 @@
...
+        aws
+      elsif @platform.name == 'azure'
+        azure
+      else
+        if @platform[:uuid_command]
+          result = @backend.run_command(@platform[:uuid_command])
+          return uuid_from_string(result.stdout.chomp) if result.exit_status == 0 && !result.stdout.empty?
+        end
+
+        raise 'Could not fine platform uuid! please set a uuid_command for your platform'
```
TrevorBramble: `s/fine/find/` =^)

lib/train/platforms/uuid.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    def aws
+      # parse the aws account id
+      client = @backend.aws_client(Aws::IAM::Client)
+      arn = client.get_user.user.arn
+      uuid_from_string(arn.split(':')[4])
+    end
+
+    def azure
+      # use subscription_id or tenant_id
```
TrevorBramble: Redundant comment, yeah? Also the kind that becomes a lie when the implementation changes.

lib/train/platforms/uuid.rb
```diff
@@ -0,0 +1,135 @@
...
+    end
+
+    def find_or_create_uuid
+      if @platform.unix?
+        unix
+      elsif @platform.windows?
+        windows
+      elsif @platform.name == 'aws'
+        aws
+      elsif @platform.name == 'azure'
```
TrevorBramble: Why aren't there predicate methods for aws and azure already?

lib/train/platforms/uuid.rb
```diff
@@ -0,0 +1,135 @@
...
+      result = @backend.run_command("'#{uuid}' | Out-File -encoding ASCII #{ENV['SYSTEMROOT']}\\machine-uuid")
+      if result.exit_status !=0
+        local_uuid_path = "#{ENV['HOMEDRIVE']}#{ENV['HOMEPATH']}\\.system\\machine-uuid"
+        warn "Cannot write uuid to `#{ENV['SYSTEMROOT']}\\machine-uuid`, falling back to #{local_uuid_path} instead"
+        # fallback to user location
+        result = @backend.run_command("new-item -force -path #{local_uuid_path} -value '#{uuid}' -type file")
+        raise "Cannot write uuid to `#{local_uuid_path}\\machine-uuid`" if result.exit_status !=0
+      end
+      uuid
+    end
```
TrevorBramble: I see about 5 methods lurking inside this one. Is that sort of abstraction unconventional in this project?

jquick: yeah at this point they should be divided up as sub private methods. Thanks for this, ill grab it when I move them over.

lib/train/platforms/uuid.rb
jquick: due to <name>? only getting created for platform families. I have been thinking about this and I will wrap aws/azure in their own families so we get them.

TrevorBramble: Gotcha. Yeah it'd be great to have the uniform api. Then the `case` statement could be collapsed. =^)

lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -87,5 +87,65 @@ def cisco_show_version
...
+    end
+
+    def unix_write_uuid
+      uuid = SecureRandom.uuid
+      result = @backend.run_command("echo \"#{uuid}\" > /etc/machine-uuid")
+      unless result.exit_status.zero?
+        # fallback to user location
+        warn 'Cannot write uuid to `/etc/machine-uuid`, falling back to ~/.local/etc/machine-uuid instead.'
+        result = @backend.run_command("mkdir -p #{ENV['HOME']}/.local/etc; echo \"#{uuid}\" > #{ENV['HOME']}/.local/etc/machine-uuid")
+        raise 'Cannot write uuid to `~/.local/etc/machine-uuid`.' unless result.exit_status.zero?
```
jerryaldrichiii: Since we surround `/etc/machine-uuid` with backticks, can we do the same for `~/.local/etc/machine-uuid` as well? Also, can we interpolate this in some way so that the home directory is expanded and shown to the user?

lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -87,5 +87,65 @@ def cisco_show_version
...
+      uuid
+    end
+
+    def uuid_from_command
+      return unless @platform[:uuid_command]
+      result = @backend.run_command(@platform[:uuid_command])
+      uuid_from_string(result.stdout.chomp) if result.exit_status.zero? && !result.stdout.empty?
+    end
+
+    def uuid_from_string(string)
```
jerryaldrichiii: Could we add a few comments explaining the conversion here?

lib/train/platforms/detect/helpers/os_windows.rb
```diff
@@ -75,5 +75,59 @@ def read_wmic_cpu
...
+      result.stdout.chomp
+    end
+
+    def windows_write_uuid
+      uuid = SecureRandom.uuid
+      result = @backend.run_command("'#{uuid}' | Out-File -encoding ASCII #{ENV['SYSTEMROOT']}\\machine-uuid")
+      unless result.exit_status.zero?
+        # fallback to user location
+        local_uuid_path = "#{ENV['HOMEDRIVE']}#{ENV['HOMEPATH']}\\.system\\machine-uuid"
+        warn "Cannot write uuid to `#{ENV['SYSTEMROOT']}\\machine-uuid`, falling back to #{local_uuid_path} instead."
```
jerryaldrichiii: Can we surround `#{local_uuid_path}` with backticks?

jquick:
Do not merge yet, need to add the new https://github.com/chef/chef/pull/7016/files chef_guid

jquick:
Added, should be good to go!

PR #271, opened 2018-03-22
Add transport for Cisco IOS

jerryaldrichiii:
This adds transport support for Cisco IOS devices.

Given that the IOS SSH server closes the SSH channel after the first read (see: https://github.com/net-ssh/net-ssh/issues/24), it is required that we ensure the SSH channel stays open.

This is done here by creating a channel, creating a buffer to receive data/manipulate that data on that channel, and requesting a shell. This, coupled with the handling of the lack of exit codes and providing a method for escalating to enable mode, should allow support for Cisco IOS devices.

### DIFF
--- a/lib/train/transports/cisco_ios.rb
+++ b/lib/train/transports/cisco_ios.rb
@@ -19,8 +19,7 @@ module Train::Transports
     option :enable_password
 
     def connection
-      validate_options(@options)
-      @connection ||= Connection.new(@options)
+      @connection ||= Connection.new(validate_options(@options).options)
     end
 
     class Connection < BaseConnection
--- a/lib/train/transports/cisco_ios.rb
+++ b/lib/train/transports/cisco_ios.rb
@@ -27,10 +27,13 @@ module Train::Transports
       def initialize(options)
         super(options)
 
+        @session = nil
+        @buf = nil
+
+        # Delete options to avoid passing them in to `Net::SSH.start` later
         @host = @options.delete(:host)
         @user = @options.delete(:user)
         @port = @options.delete(:port)
-
         @enable_password = @options.delete(:enable_password)
 
         @prompt = /^\S+[>#]\r\n.*$/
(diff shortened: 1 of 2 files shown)


lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,135 @@
...
+
+    # Used to elevate to enable mode (similar to `sudo su` in Linux)
+    option :enable_password
+
+    def connection
+      validate_options(@options)
+      @connection ||= Connection.new(@options)
+    end
+
+    def validate_options(options)
```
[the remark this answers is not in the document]
jerryaldrichiii: Nope, you're right. Will remove.

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,135 @@
...
+      end
+
+      def run_command_via_connection(cmd)
+        @session ||= establish_connection
+
+        result = run_command_via_channel(cmd)
+        CommandResult.new(*format_result(result))
+      end
+
+      def format_result(result)
```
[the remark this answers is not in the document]
jerryaldrichiii: Yup, should be done in latest commit.

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,147 @@
...
+          stderr_with_exit_1
+        when /Incomplete command/
+          stderr_with_exit_1
+        when /Invalid input detected/
+          stderr_with_exit_1
+        when /Unrecognized host/
+          stderr_with_exit_1
+        else
+          stdout_with_exit_0
+        end
```
TrevorBramble: As far as better ways to organize this, it seems like they're all just string literals (`when 'Bad IP address'` and so forth would have worked too) and they all should be treated as exit 1, so this is what came to mind. Basically just extract the data from the logical flow.

```ruby
ERROR_STRINGS = [
  'Bad IP address',
  'Incomplete command',
  'Invalid input detected',
  'Unrecognized host'
]

# IOS commands do not have an exit code, so we must capture known errors
def format_result(result)
  stderr_with_exit_1 = ['', result, 1]
  stdout_with_exit_0 = [result, '', 0]

  ERROR_strings.include?(result) ? stderr_with_exit_1 : stdout_with_exit_0
end
```
I debated inlining the return values, but I really like those explaining variables!

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,147 @@
...
+
+      private
+
+      def establish_connection
+        logger.debug("[SSH] opening connection to #{self}")
+
+        @ssh = Net::SSH.start(
+          @host,
+          @user,
+          @options.delete_if { |_key, value| value.nil? },
```
TrevorBramble: Previously this was `@options.clone.delete_if { |_key, value| value.nil? }` and I presumed the intent was to avoid mutating `@options` even if you don't want the entries with nil values for this statement, and that's why I suggested `@options.reject { |_, value| value.nil? }` (which doesn't mutate anything at all) instead.

So I just wanted to make sure there wasn't confusion here. Maybe it's OK to strip the empty pairs here, but it seems like a thing that could cause trouble if you're not expecting it elsewhere.

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,147 @@
...
+        output.sub!(/#{cmd}\r\n/, '')
+
+        # Remove trailing prompt
+        output.gsub!(/\S+[>#](\r\n|$)/, '')
+
+        # Remove trailing returns/newlines
+        output.gsub!(/(\r\n)+$/, '')
+
+        output
+      end
```
TrevorBramble: `String#sub` and `String#gsub` return the modified string, so you can chain them. Also, giving the regular expression matchers explaining variables can help avoid an almost certain fate of the regexp being changed, but not the comment that explains it.
```ruby
def format_output(output, cmd)
  leading_prompt        = /(\r\n|^)\S+[>#]/
  command_string        = /#{cmd}\r\n/
  trailing_prompt       = /\S+[>#](\r\n|$)/
  trailing_line_endings = /(\r\n)+$/

  output
    .sub(leading_prompt, '')
    .sub(command_string, '')
    .gsub(trailing_prompt, '')
    .gsub(trailing_line_endings, '')
end
```

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,147 @@
...
+    option :port, default: 22, required: true
+
+    option :password, required: true
+
+    # Used to elevate to enable mode (similar to `sudo su` in Linux)
+    option :enable_password
+
+    def connection
+      validate_options(@options)
+      @connection ||= Connection.new(@options)
```
TrevorBramble: Do you actually want to validate the options every time or should this instead return an existing connection _or_ validate and connect?

Oh, I looked up `validate_options`; it returns the validated options, so you can pass it right through. =^)
```ruby
def connection
  @connection ||= Connection.new(validate_options(@options))
end
```

lib/train/transports/cisco_ios.rb
jerryaldrichiii: Nope, I see what you mean now, I'll go with `reject` here.

lib/train/transports/cisco_ios.rb
jerryaldrichiii: I was unable to get that to work. Looks like it adds some other stuff in the return object

See: https://gist.github.com/00c569cfe7ac85ff6eb397f8d74b7a3d

I could do: 
```
@connection ||= Connection.new(validate_options(@options).options)
```
I pushed that up in the latest, commit. Let me know if that is incorrect in some way. I definitely only want to call it if `@connection` doesn't exist.

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,141 @@
...
+    def connection
+      @connection ||= Connection.new(validate_options(@options).options)
+    end
+
+    class Connection < BaseConnection
+      def initialize(options)
+        super(options)
+
+        @session = nil
+        @buf = nil
```
TrevorBramble: [nit] Is there a reason for these? In Ruby, instance variables are `nil` by default, and attempting to access the value of one without having declared it already just returns `nil`. `@session` should only ever be assigned or accessed via `session()`, even. `@buf` may be shared by those two methods, but `run_command_via_connection()` clears it on each invocation anyway.

lib/train/transports/cisco_ios.rb
```diff
@@ -0,0 +1,141 @@
...
+      ERROR_MATCHERS = [
+        'Bad IP address',
+        'Incomplete command',
+        'Invalid input detected',
+        'Unrecognized host',
+      ].freeze
+
+      def format_result(result)
+        stderr_with_exit_1 = ['', result, 1]
+        stdout_with_exit_0 = [result, '', 0]
```
TrevorBramble: I put my finger on why these bother me. One of them is always a spurious assignment. At this point I'd definitely either extract them as methods (`format_err(result)`, `format_out(result)`) or inline the arrays, or inline `CommandResult.new()`.

```
def format_result(result)
  if ERROR_MATCHERS.none? { |e| result.include?(e) }
    CommandResult.new(result, '', 0)
  else
    CommandResult.new('', result, 1)
  end
end
```
I do wish there were `CommandResult` subclasses that revealed intent though. `ErrorResult.new(result)` which automatically assigns `result` to stderr and returns 1, etc.

jerryaldrichiii: I'm glad you put a finger on it. I took your code suggestion and ran with it. Let me know if it's any better.

lib/train/transports/cisco_ios.rb
jerryaldrichiii: Nope, you make a solid argument. Killing those with fire.

PR #275, opened 2018-03-28
small fix to make sure windows directories with the archive bit set a…

devoptimist:
…re reported as a directory and not a file
Signed-off-by: Steve Brown <sbrown@chef.io>


jquick:
Confirmed commit is signed, DCO bot having issues.

PR #277, opened 2018-04-02
Modify IOS connection options and add handling for commands with pipes

jerryaldrichiii:
This does the following:
  - Modifies the connection options passed to `Net::SSH.start` to match the other SSH implementation
  - Adds handling for commands containing `|`


jerryaldrichiii:
Decided to pull the handling for `show run` I think that should be handled by the user consuming Train.

jerryaldrichiii:
Closing this in favor of a PR that reimplements the Cisco IOS connection under the SSH transport.

See: https://github.com/chef/inspec/pull/2905

PR #278, opened 2018-04-03
Initialize FileManager using '@service'

marcparadise:
`FileManager` expects a `Connection` instance and not a
`Shell` instance, but `#session` returns a `Shell`.
This change passes the Connection instance into `WinRM::FileManager.new` instead of the shell. 


marcparadise:
Thanks @jquick , I've made that change.

PR #283, opened 2018-04-16
Initial import of transport for GCP.

skpaterson:
Add transport for Google Cloud Platform.  Note, this introduces a gem dependency issue with berkshelf.

### DIFF
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -7,7 +7,6 @@ require 'google/apis/compute_v1'
 require 'google/apis/storage_v1'
 require 'google/apis/iam_v1'
 require 'googleauth'
-require 'JSON'
 
 module Train::Transports
   class Gcp < Train.plugin(1)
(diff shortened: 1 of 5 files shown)


clintoncwolfe:
Refs https://github.com/chef/inspec/issues/2722

lib/train/transports/gcp.rb
```diff
@@ -0,0 +1,101 @@
...
+
+        connect
+      end
+
+      def platform
+        direct_platform('gcp', @platform_details)
+      end
+
+      # Instantiate some named classes for ease of use
+      def gcp_compute_client
```
jquick: if your going to make helper methods can you use the `gcp_client` for them all? That way the logic is in one spot.
```
def gcp_iam_client
   gcp_client(Google::Apis::IamV1::IamService)
end
```

skpaterson: Good shout! I'll update that now.

PR #285, opened 2018-04-17
Add UUID for Cisco IOS and Nexus devices

jerryaldrichiii:
This adds a UUID command for Nexus devices and a way to extract a UUID from IOS devices.

### DIFF
--- a/lib/train/transports/cisco_ios_connection.rb
+++ b/lib/train/transports/cisco_ios_connection.rb
@@ -23,6 +23,13 @@ class Train::Transports::SSH
       "ssh://#{@user}@#{@host}:#{@port}"
     end
 
+    def unique_identifier
+      result = run_command_via_connection('show inventory').stdout
+      result.split("\r\n\r\n").each do |section|
+        return section.split('SN: ')[1].strip if section.include?('Chassis')
+      end
+    end
+
     private
 
     def establish_connection
--- a/test/unit/transports/cisco_ios_connection.rb
+++ b/test/unit/transports/cisco_ios_connection.rb
@@ -31,6 +31,16 @@ describe 'CiscoIOSConnection' do
     end
   end
 
+  describe '#unique_identifier' do
+    it 'returns the correct identifier' do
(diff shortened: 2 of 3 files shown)


test/unit/transports/cisco_ios_connection.rb
```diff
@@ -31,6 +31,16 @@
     end
   end
 
+  describe '#unique_identifier' do
+    it 'returns the correct identifier' do
+      output = "NAME: \"Chassis\", DESCR: \"Cisco 7206VXR, 6-slot chassis\"\r\nPID: CISCO7206VXR      , VID:    , SN: 4279256517 \r\n\r\nNAME: \"NPE400 0\", DESCR: \"Cisco 7200VXR Network Processing Engine NPE-400\"\r\nPID: NPE-400           , VID:    , SN: 11111111   \r\n\r\nNAME: \"module 0\", DESCR: \"I/O FastEthernet (TX-ISL)\"\r\nPID: C7200-IO-FE-MII/RJ45=, VID:    , SN: 4294967295 \r\n\r\nNAME: \"Power Supply 1\", DESCR: \"Cisco 7200 AC Power Supply\"\r\nPID: PWR-7200-AC       , VID:    , SN:            \r\n\r\nNAME: \"Power Supply 2\", DESCR: \"Cisco 7200 AC Power Supply\"\r\nPID: PWR-7200-AC       , VID:    , SN:            "
+      Train::Transports::SSH::CiscoIOSConnection.any_instance
+        .expects(:run_command_via_connection).with('show inventory')
+        .returns(OpenStruct.new(stdout: output))
+      connection.unique_identifier.must_equal('4279256517')
```
chris-rock: I think we need to use uuid v5 to generate a uuid from a sha. otherwise we do not have consistent formats for the uuid

jquick: This happens automatically upstream here:
https://github.com/chef/train/blob/master/lib/train/platforms/detect/uuid.rb#L19

lib/train/transports/cisco_ios_connection.rb
```diff
@@ -23,6 +23,13 @@ def uri
       "ssh://#{@user}@#{@host}:#{@port}"
     end
 
+    def unique_identifier
+      result = run_command_via_connection('show inventory').stdout
+      result.split("\r\n\r\n").each do |section|
+        return section.split('SN: ')[1].strip if section.include?('Chassis')
```
jquick: Is it possible for Chassie not to be set? If we have multiple boxes like this we will end up having the same uuid of nil on them. If it is possible to be nil we should fall back to a MAC or something.

test/unit/transports/cisco_ios_connection.rb
chris-rock: Ah nice!

lib/train/transports/cisco_ios_connection.rb
jerryaldrichiii: I don't think it is possible for Chassis not to be set. We can't fall back to MAC because modules can change. Also, I don't know how to decided which interface to choose. Example: FastEthernet vs GigabitEthernet.

jquick: If its always set we should be good. Thanks!

PR #293, opened 2018-04-30
Split train into a core gem.

miah:
Provide a `train-core.gemspec` to allow distribution without additional backend dependencies.

Signed-off-by: Miah Johnson <miah@chia-pet.org>

### DIFF
--- a/train-core.gemspec
+++ b/train-core.gemspec
@@ -1,0 +1,34 @@
+# encoding: utf-8
+lib = File.expand_path('../lib', __FILE__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
+require 'train/version'
+
+Gem::Specification.new do |spec|
+  spec.name          = 'train-core'
+  spec.version       = Train::VERSION
+  spec.authors       = ['Dominik Richter']
+  spec.email         = ['dominik.richter@gmail.com']
+  spec.summary       = 'Transport interface to talk to a selected set of backends.'
+  spec.description   = 'A minimal Train with a selected set of backends, ssh, winrm, and docker.'
+  spec.homepage      = 'https://github.com/chef/train/'
+  spec.license       = 'Apache-2.0'
+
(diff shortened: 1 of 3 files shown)


train-core.gemspec
```diff
@@ -0,0 +1,34 @@
...
+  spec.authors       = ['Dominik Richter']
+  spec.email         = ['dominik.richter@gmail.com']
+  spec.summary       = 'Transport interface to talk to a selected set of backends.'
+  spec.description   = 'A minimal Train with a selected set of backends, ssh, winrm, and docker.'
+  spec.homepage      = 'https://github.com/chef/train/'
+  spec.license       = 'Apache-2.0'
+
+  spec.files = %w{train-core.gemspec README.md LICENSE Gemfile CHANGELOG.md} + Dir
+               .glob('{lib,test}/**/*', File::FNM_DOTMATCH)
+               .reject { |f| File.directory?(f) || f =~ /aws|azure|gcp/ }
```
jquick: I think for the train-core gem we want it to only have the `local` transport. So we will want to reject all the transports except for that one. You can probably nix all the spec/tests as well.

jquick:
Built and tested without issues:
```
jquick@Jareds-MBP:~/Chef/train/_test
(miah/train_core)> ruby aws.rb
/Users/jquick/.gem/ruby/2.4.3/gems/train-core-1.4.1/lib/train.rb:47:in `rescue in load_transport': Can't find train plugin "aws". Please install it first. (Train::UserError)
        from /Users/jquick/.gem/ruby/2.4.3/gems/train-core-1.4.1/lib/train.rb:39:in `load_transport'
        from /Users/jquick/.gem/ruby/2.4.3/gems/train-core-1.4.1/lib/train.rb:20:in `create'
        from aws.rb:7:in `<main>'
jquick@Jareds-MBP:~/Chef/train/_test
(miah/train_core)> ruby local.rb
"darwin"
"mac_os_x"
["darwin", "bsd", "unix", "os"]
```

PR #297, opened 2018-05-08
Allow nil password and www_form_encoded_password to work together.

marcparadise:
This corrects a bug previously untested path in the change to support www-form-encoded password . 

When a URL does not contain a password but options contains
www_form_encoded_password: true, an exception was raised
because we can't decode a nil value.


marcparadise:
np - sorry for the bug :D

PR #305, opened 2018-05-30
Upgrade aws-sdk dependency to version ~>3 .

cliffano:
Since aws-sdk is now at version 3.0.1 , I upgraded the dependency in gemspec to `~> 3` .
I bundle installed the dependencies and ran `rake` to ensure it passed the build locally.

According to upgrade doco https://aws.amazon.com/blogs/developer/upgrading-from-version-2-to-version-3-of-the-aws-sdk-for-ruby-2/ , it's as simple as changing `~> 2` to `~> 3` since train is requiring aws-sdk and not aws-sdk-core.

### DIFF
--- a/train.gemspec
+++ b/train.gemspec
@@ -33,7 +33,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'winrm', '~> 2.0'
   spec.add_dependency 'winrm-fs', '~> 1.0'
   spec.add_dependency 'docker-api', '~> 1.26'
-  spec.add_dependency 'aws-sdk', '~> 2'
+  spec.add_dependency 'aws-sdk', '~> 3'
   spec.add_dependency 'azure_mgmt_resources', '~> 0.15'
   spec.add_dependency 'google-api-client', '~> 0.19.8'
   spec.add_dependency 'googleauth', '~> 0.6.2'


cliffano:
The Travis builds failed on ruby-2.2.7 and ruby-2.3.4 due to `RuntimeError: Unable to read the credential file specified by GOOGLE_APPLICATION_CREDENTIALS: file` which seems to be caused by ` [SSH] connection failed, terminating (#<Errno::ECONNREFUSED: Connection refused - connect(2) for 127.0.0.1:1>)` .

The SSH connection is fine on the other builds with ruby-2.4.1 .

Could the connection failure be intermittent?

clintoncwolfe:
Hi @cliffano , 
  Thanks for the PR!  Unfortunately, we are unable to make the change to SDK 3 at this time.  Because 2 and 3 cannot be used in the same ruby process, and we have a few places where that happens, we at Chef need to coordinate the upgrade across a few products.  We're aware of the issue, and as nothing as yet has compelled us to upgrade, we are currently waiting.

Refs https://github.com/inspec/inspec/issues/2571

cliffano:
@clintoncwolfe Thanks for looking this up.

How about bumping up the major version of train with the aws-sdk upgrade?
That way other Chef libraries won't get affected immediately.
I was actually hoping that train would be upgraded first, then I would create PR for the other libraries, specially InSpec.

The problem I'm encountering is like what you described about ruby 2 and 3 on the same process, but I'm on the opposite end, i.e. I have a number of other libraries that I need to use which already upgraded to aws-sdk 3, but InSpec's (through train) dependency to aws-sdk 2 is causing a conflict.

Sad to say the workaround for us is to downgrade InSpec to 1.51.6, which happened to have a (arguably incorrectly) relax aws-sdk version dependency.

PR #306, opened 2018-05-31
Don't double-escape paths

voroniys:
Signed-off-by: Stanislav Voroniy <stas@voroniy.com>
As argument to grep original path should be used and not escaped, since it is already included in single quotes.


clintoncwolfe:
Refs https://github.com/inspec/inspec/issues/3098

jquick:
closing and reopening to refresh tests.

PR #309, opened 2018-06-23
Add shallow_link_path to inspect symlink direct link

ColinHebert:
Currently symlinks are read recursively.
There is the case where people simply want to inspect what the current symlink points to and not resolve the entire symlink chain. This `shallow_link_path` gives access to that information.

Helps fixing inspec/inspec#3100 and inspec/inspec#2106

### DIFF
--- a/lib/train/file/remote/aix.rb
+++ b/lib/train/file/remote/aix.rb
@@ -12,6 +12,12 @@ module Train
             @backend.run_command("perl -e 'print readlink shift' #{@spath}").stdout.chomp
         end
 
+        def shallow_link_path
+          return nil unless symlink?
+          @shallow_link_path ||=
+            @backend.run_command("perl -e 'print readlink shift' #{@spath}").stdout.chomp
+        end
+
         def mounted
           @mounted ||= @backend.run_command("lsfs -c #{@spath}")
         end
--- a/lib/train/file/remote/windows.rb
+++ b/lib/train/file/remote/windows.rb
@@ -77,6 +77,10 @@ module Train
           nil
         end
 
+        def shallow_link_path
+          nil
+        end
+
         def mounted
           nil
         end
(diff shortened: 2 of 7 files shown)


lib/train/file/remote/windows.rb
```diff
@@ -77,6 +77,10 @@ def link_path
           nil
         end
 
+        def shallow_link_path
```
clintoncwolfe: I'm no expert, but my understanding is that Win10 does support symbolic links.

ColinHebert: Indeed, that's why we have a nil (same as link_path above)

ColinHebert:
@clintoncwolfe I'm reusing the same pattern as `link_path` which has the same issue.

lib/train/file/remote/aix.rb
```diff
@@ -12,6 +12,12 @@ def link_path
             @backend.run_command("perl -e 'print readlink shift' #{@spath}").stdout.chomp
         end
 
+        def shallow_link_path
```
jquick: Should we just alias this if its going to be exactly the same as `link_path`?

ColinHebert: I don't really want to do that. Turns out on aix, `link_path` doesn't respect the spec (which is doing a recursive lookup), this is a problem that should be fixed independently

jquick:
Closing and re-opening to trigger tests.

lib/train/file/remote/windows.rb
clintoncwolfe: What I mean to say is that because Win10 supports symbolic links, I would expect the implementation of link_path and shallow_link_path to be able to handle a symbolic link on windows 10, and return nil on other versions of windows that do not support symlinks.

jquick:
I have created a issue for the windows 10 additional support. We can merge this as is and focus on additional enhancement later.

Issue #163, opened 2016-11-04
Bastion support for SSH transport


jamtur01:
The SSH transport should support bastion or jump hosts. A lot of folks run their hosts, especially in AWS/GCP, behind bastion or jump hosts. Without this support it's hard for downstream tools like InSpec to run tests.

thiagonache:
It's important to be able to configure bastion without use ssh_config workaround for automated tests use cases.

aaronlippold:
@arlimus @chris-rock This will be a key support feature for many of your government users.

darksheer:
Any update on this as it's blocking a commonly used deployment paths in CSP's such as AWS VPC private networks?

clintoncwolfe:
I believe this was fixed by #310 - please re-open if not, thanks!

PR #310, opened 2018-06-25
Adding proper bastion support

frezbo:
This PR addresses these issues:
 - #163 

If the skeleton is good, I could raise PR's that reference these:
- https://github.com/inspec/inspec/issues/431
- https://github.com/inspec/kitchen-inspec/pull/176

@clintoncwolfe @jquick @aaronlippold @chris-rock  Any thoughts or suggestions?

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -58,6 +58,10 @@ module Train::Transports
     option :max_wait_until_ready, default: 600
     option :compression, default: false
     option :pty, default: false
+    option :custom_proxy_command, default: nil
+    option :bastion_host, default: nil
+    option :bastion_user, default: 'root'
+    option :bastion_port, default: 22
 
     option :compression_level do |opts|
       # on nil or false: set compression level to 0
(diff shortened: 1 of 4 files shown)


lib/train/transports/ssh.rb
```diff
@@ -150,7 +158,10 @@ def connection_options(opts)
         keys:                   opts[:key_files],
         password:               opts[:password],
         forward_agent:          opts[:forward_agent],
-        proxy_command:          opts[:proxy_command],
+        custom_proxy_command:   opts[:custom_proxy_command],
```
jquick: I think this will break any current `proxy_command` usage. You could probably alias or set `custom_proxy_command` from `proxy_command` if set.

frezbo: @jquick I can't seem to think why I used a `custom` prefix then. Good suggestion. Fixed and pushed. Thanks for the suggestion.

frezbo:
The failure seems unrelated to my change.

frezbo:
@jquick  Is there a new release being scheduled? I would like to add the same support to inspec and bump up the train version. Is there any place we can discuss this. Inspec slack? (Ironically it's still down ATM) :+1:

jquick:
We will release train Thursday, ill take care of getting it pined up! Thanks @frezbo! I am also around the community chef slack when it up :D.

frezbo:
Thanks @jquick time to start on inspec.

PR #317, opened 2018-06-29
Remove the delivery cookbook

tas50:
We're not pushing this project through workflow so there's no need for
this.


tas50:
Ping @jquick

PR #318, opened 2018-06-29
Update URLs to inspec/train

tas50:
This should be most of them. This gets the badges working again.


tas50:
Ping @jquick

PR #321, opened 2018-07-06
Add VMware transport

jerryaldrichiii:
This adds the VMware transport via PowerCLI.

See: https://github.com/inspec/inspec/issues/3133

### DIFF
--- a/lib/train/platforms/detect/specifications/api.rb
+++ b/lib/train/platforms/detect/specifications/api.rb
@@ -11,6 +11,8 @@ module Train::Platforms::Detect::Specifications
       plat.name('aws').in_family('cloud')
       plat.name('azure').in_family('cloud')
       plat.name('gcp').in_family('cloud')
+      plat.name('vmware').in_family('cloud')
+
       plat.family('iaas').in_family('api')
       plat.name('oneview').in_family('iaas')
     end
--- a/lib/train/transports/vmware.rb
+++ b/lib/train/transports/vmware.rb
@@ -1,0 +1,190 @@
+# encoding: utf-8
+require 'train/plugins'
+require 'open3'
+require 'ostruct'
+require 'json'
+require 'mkmf'
+
+module Train::Transports
+  class VMware < Train.plugin(1)
+    name 'vmware'
+    option :viserver, default: ENV['VISERVER']
+    option :username, default: ENV['VISERVER_USERNAME']
+    option :password, default: ENV['VISERVER_PASSWORD']
+    option :insecure, default: false
+
+    def connection(_ = nil)
+      @connection ||= Connection.new(@options)
+    end
+
+    class Connection < BaseConnection # rubocop:disable ClassLength
+      POWERSHELL_PROMPT_REGEX = /PS\s.*> $/
+
+      def initialize(options)
+        super(options)
+
+        options[:viserver] = options[:viserver] || options[:host]
+        options[:username] = options[:username] || options[:user]
+
+        @username = options[:username]
+        @viserver = options[:viserver]
+        @session = nil
+        @stdout_buffer = ''
+        @stderr_buffer = ''
+
+        @powershell_binary = detect_powershell_binary
+
+        if @powershell_binary == :powershell
+          require 'train/transports/local'
+          @powershell = Train::Transports::Local::Connection.new(options)
+        end
+
+        if options[:insecure] == true
+          run_command_via_connection('Set-PowerCLIConfiguration -InvalidCertificateAction Ignore -Scope Session -Confirm:$False')
+        end
+
+        @platform_details = {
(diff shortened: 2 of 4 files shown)


lib/train/transports/vmware.rb
```diff
@@ -0,0 +1,190 @@
...
+      @connection ||= Connection.new(@options)
+    end
+
+    class Connection < BaseConnection # rubocop:disable ClassLength
+      POWERSHELL_PROMPT_REGEX = /PS\s.*> $/
+
+      def initialize(options)
+        super(options)
+
+        options[:viserver] = options[:viserver] || options[:host]
```
clintoncwolfe: Fun shorthand: try `||=`.  Like `+=`, but for or.  It's great for setting defaults.
```
options[:viserver] ||= options[:host]
     
```

lib/train/transports/vmware.rb
```diff
@@ -0,0 +1,190 @@
...
+          else
+            message += result.stderr.gsub(/-Password .*\s/, '-Password REDACTED')
+          end
+
+          raise message
+        end
+      end
+
+      def local?
+        true
```
clintoncwolfe: Is this constant?

lib/train/transports/vmware.rb
```diff
@@ -0,0 +1,190 @@
+# encoding: utf-8
```
clintoncwolfe: There is a lot going on here that I really wish we had integration tests for.  The unit tests are solid - I believe that the accessors and option parsers work.  I'm sure info about integration testing is coming, likely in the resource pack.

clintoncwolfe: That said, pleas update the README.md to include vmware, and provide configuration information there.

lib/train/platforms/detect/specifications/api.rb
```diff
@@ -11,6 +11,8 @@ def self.load
       plat.name('aws').in_family('cloud')
       plat.name('azure').in_family('cloud')
       plat.name('gcp').in_family('cloud')
+      plat.name('vmware').in_family('cloud')
```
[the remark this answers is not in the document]
jerryaldrichiii: Yup, I struggled with this as well. After discussing, @jquick and @chris-rock decided that is should be under `cloud`.

lib/train/transports/vmware.rb
jerryaldrichiii: Yup, I too would like integration tests, I'm just not quite sure how to approach it here.

jerryaldrichiii: I've added sections covering basic usage. Detailed info (such as PWSH/PowerCLI installation will be in the resource pack).

Does that work for you? 😄

lib/train/transports/vmware.rb
jerryaldrichiii: Indeed it is, will do!

lib/train/transports/vmware.rb
jerryaldrichiii: I think so, I'm trying to override: https://github.com/inspec/train/blob/d7b958c4cd1ee3b7993bd0fc05a85fb25544046c/lib/train/plugins/base_connection.rb#L76

test/unit/transports/vmware_test.rb
```diff
@@ -0,0 +1,165 @@
...
+    end
+
+    it 'uses the Local connection when Windows PowerShell is found' do
+      require 'train/transports/local'
+      Train::Transports::Local::Connection.expects(:new)
+      create_transport(
+        stub_options: {
+          powershell_binary: :powershell
+        }
+      ).connection
```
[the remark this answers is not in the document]
jerryaldrichiii: Well, I expect it to create a `Local::Connection` if the PowerShell binary is proper PowerShell (vs `pwsh`).

So I did an `expect`, then created the connection. Is there a better way?

test/unit/transports/vmware_test.rb
```diff
@@ -0,0 +1,165 @@
...
+    it 'raises certificate error when stderr matches regular expression' do
+      e = proc {
+        create_transport(
+          stub_options: {
+            mock_connect_result: mock_connect_result(
+              'Invalid server certificate', 1
+            )
+          }
+        ).connection
+      }.must_raise(RuntimeError)
```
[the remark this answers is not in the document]
jerryaldrichiii: Hmmm...should I raise a different error? My plan was to ensure it raises, capture the exception in `e` then make sure the message is what I expect.

test/unit/transports/vmware_test.rb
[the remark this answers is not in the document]
clintoncwolfe: Ahh, I see.  the `expects` should fire.  Too many testing paradigms on this project :-)

PR #324, opened 2018-07-18
Pulls file credentials parsing out of Azure class

dmccown:
This change pulls file credentials parsing out of Azure.rb. In
extracting the file credentials parsing I refactored the code to
introduce a strategy pattern around the different ways one may access
credentials from the file.

I introduced more unit testing for the extracted classes based on their
public interfaces.

This change also fixes an implementation error where asking for the
credential at index 0, would return the last item in your credentials
file. It will now raise an error that you should give a value greater
than 0. Credit to @Padgett for catching this bug!


lib/train/transports/azure.rb
```diff
@@ -38,7 +38,8 @@ def initialize(options)
         @cache[:api_call] = {}
 
         if @options[:client_secret].nil? && @options[:client_id].nil?
-          parse_credentials_file
+          values = AzureHelpers::FileCredentials.parse(@options)
+          @options.merge!(values)
```
TrevorBramble: I'd skip the local variable and inline this if there's no conditional usage of it.

lib/train/transports/azure_helpers/file_credentials.rb
```diff
@@ -0,0 +1,39 @@
...
+            FileParser.new(credentials)
+          end
+        end
+
+        def self.creds(subscription_id, credentials)
+          {
+            subscription_id: subscription_id,
+            tenant_id:       credentials[subscription_id]['tenant_id'],
+            client_id:       credentials[subscription_id]['client_id'],
+            client_secret:   credentials[subscription_id]['client_secret']
```
TrevorBramble: Either these should be `fetch` operations or there should be a `rescue NoMethodError` statement for trying to call `[]` on `nil`. Unless you're *really sure* all those values exist before getting here. =^)

lib/train/transports/azure_helpers/file_parser.rb
```diff
@@ -0,0 +1,23 @@
...
+        def initialize(credentials)
+          @credentials = credentials
+
+          validate!
+        end
+
+        def validate!
+          if @credentials.sections.count > 1
+            raise 'Multiple credentials detected, please set the AZURE_SUBSCRIPTION_ID environment variable.'
+          end
```
TrevorBramble: Personally I find using guard statements easier to scan.
```ruby
def validate!
  return if @credentials.section.count < 2

  raise 'Multiple credentials detected, please set the AZURE_SUBSCRIPTION_ID environment variable.'
end
```

test/unit/transports/azure_helpers/file_credentials_test.rb
```diff
@@ -0,0 +1,115 @@
...
+
+    result = Train::Transports::AzureHelpers::FileCredentials.parse(options)
+
+    assert_equal('my_tenant_id',       result[:tenant_id])
+    assert_equal('my_client_id',       result[:client_id])
+    assert_equal('my_client_secret',   result[:client_secret])
+    assert_equal('my_subscription_id', result[:subscription_id])
+  end
+
+  it 'raises and error when no subscription id given and multiple entries' do
```
TrevorBramble: "raises and error"

lib/train/transports/azure_helpers/file_credentials.rb
dmccown: `credentials[subscription_id]` should be safe. If we use either the 0 index method or the index based lookup the subscription id is read from the credentials file. If find the credentials with the subscription_id then we have a guard to ensure that section exists in the file. 

There may be an edge case if the file has no entries. I can test that out.

dmccown:
I found a bug in this code. Please hold off merging while I add a test and fix the bug.

PR #327, opened 2018-07-26
Adds connection to Graph RBAC API 

r-fennell:


jquick:
We will have to confirm all the upstream inspec artifact builds are ok with 'azure_graph_rbac'. To do this create a inspec branch based off this train branch and then create a ad-hoc inspec build in our pipeline.

lib/train/transports/azure.rb
```diff
@@ -54,14 +56,21 @@ def platform
         force_platform!('azure', @platform_details)
       end
 
+      def graph_client
```
jquick: This may be confusing as the main client object for all azure clients is under this. I think it may be clearer to just put some logic into the `azure_client` method to pick out the graph client. This way we don't end up with `inspec.backend.graph_client` in inspec, which is a type of `azure_client`.

r-fennell: I did initially have have this logic within the `azure_client` method, however @dmccown and I thought this convenience method was clearer. We do indeed then have `inspec.backend.graph_client` present in Inspec [here](https://github.com/inspec/inspec-azure/blob/dmccown/replace-connection-with-train/libraries/azurerm_resource.rb#L29) - is this bad?

dmccown: After talking this through, we'll change `azure_client` to handle the case that a client passes in `::Azure::GraphRbac::Profiles::Latest::Client`. The returned client will be fully hydrated and ready for consumption. So essentially it will fix an existing bug in cases you asked for a Graph client and the returned client didn't work.

jquick:
This is looking pretty good but you have some failing tests.

lib/train/transports/azure.rb
```diff
@@ -3,9 +3,11 @@
 require 'train/plugins'
 require 'ms_rest_azure'
 require 'azure_mgmt_resources'
+require 'azure_graph_rbac'
 require 'socket'
 require 'timeout'
 require 'train/transports/helpers/azure/file_credentials'
+require_relative 'clients/azure/graph_rbac'
```
dmccown: Should we just use `require` here?

lib/train/transports/azure.rb
```diff
@@ -55,13 +57,27 @@ def platform
       end
 
       def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client)
-        return klass.new(@credentials) unless cache_enabled?(:api_call)
+        # Return early if we can
```
dmccown: I'd remove the comment.

Issue #328, opened 2018-07-27
`google-cloud` Dependency is not necessary


chrisUsick:
### Description

By having train depend on `google-cloud` it makes libraries that depend on `train` have a much longer dependency resolution step when doing a bundle install. I don't think that it is necessary for `train` to depend on google-cloud gem, rather it could require only the specific dependencies that it needs. 

### Train and Platform Version

All platforms and versions.

### Replication Case

1. Create new directory
2. Create Gemfile with following content

```
source 'https://rubygems.org'
gem 'train'
```

3. run `time bundle update`

### Possible Solutions

I'm completely unfamiliar with train, I noticed this problem when debugging long dep resolution times for a chef cookbook that required `kitchen-inspec`. That said, in [lib/train/transports/gcp.rb](https://github.com/inspec/train/blob/340267d360888b363d4d966c8b6d450780679f07/lib/train/transports/gcp.rb) only the following google gems are required:

- `google-api-client`
- `googleauth`

I don't even see any reference to `google-cloud` gem anywhere.

skpaterson:
You're right, for the current implementation `google-cloud` is not required,  I'll create a PR to address this.

clintoncwolfe:
Please note that we're pushing most platform-specific transports into plugins; see https://github.com/inspec/inspec/issues/3089 SSH and winrm will remain in core; GCP, AWS, and Azure will be pushed into plugins, and we're on the fence about some others, like Docker.

chrisUsick:
Sounds like a good move and thanks for keeping me updated. I think this
issue can be closed.

clintoncwolfe:
Refs #335

clintoncwolfe:
Closing per OP - if there are any questions, please feel free to re-open.

PR #329, opened 2018-07-30
Remove not needed google-cloud dependency (see #328) and correct GCP …

skpaterson:
…platform release string.

### DIFF
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -32,8 +32,8 @@ module Train::Transports
         super(options)
 
         # additional GCP platform metadata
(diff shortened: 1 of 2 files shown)


tas50:
This is a pretty critical update for us in chef/chef-dk land. We fixed up a dep that was holding back inspec 2.2 and now we're getting this massive new set of deps. If we can get this merged an a new version of train pushed up it'd be great. I'd like to get the latest inspec up in DK.

tas50:
Thanks @skpaterson @jquick

This unblocks the ChefDK release

frezbo:
:100: this would cleanup a lot of inspec-deps

PR #336, opened 2018-08-14
Add non_interactive support for SSH

marcparadise:
This passes through the :non_interactive option to Net::SSH.start
for cases where the caller does not want the net/ssh UI to display interactive
prompting.

This is mapped to BatchMode=true in the ssh command usage.

### DIFF
--- a/test/unit/transports/ssh_test.rb
+++ b/test/unit/transports/ssh_test.rb
@@ -55,6 +55,18 @@ describe 'ssh transport' do
     end
   end
 
+  describe 'ssh options' do
+    let(:ssh) { cls.new(conf) }
+    let(:connection) { ssh.connection }
+    it 'includes BatchMode when :non_interactive is set' do
+      conf[:non_interactive] = true
+      connection.ssh_opts.include?("BatchMode=yes").must_equal true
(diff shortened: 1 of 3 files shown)


marcparadise:
Note that the travis errors are also occurring on master: 

https://travis-ci.com/inspec/train/jobs/139904605

test/unit/transports/ssh_test.rb
```diff
@@ -55,6 +55,19 @@
...
 
+  describe 'ssh options' do
+    let(:ssh) { cls.new(conf) }
+    let(:connection) { ssh.connection }
+    it 'includes BatchMode when :non_interactive is set' do
+      conf[:non_interactive] = true
+      connection.ssh_opts.include?("BatchMode=yes").must_equal true
+    end
+    it 'excludes BatchMode when :non_interactive is not set' do
+      conf.delete :non_interactive
```
tyler-ball: Because `conf` is set by a `let` statement you _should_ be able to leave this line out because it is reset for every example. Could you try removing this?

Leaving it in makes these examples look like they are order dependent, which is a unit test smell IMO

PR #337, opened 2018-08-15
Fixes an issue where the credential file was nil

dmccown:
This adds a test and a guard to handle the case where the credentials
file can be nil.


lib/train/transports/helpers/azure/file_credentials.rb
```diff
@@ -12,6 +12,7 @@ class FileCredentials
         DEFAULT_FILE = ::File.join(Dir.home, '.azure', 'credentials')
 
         def self.parse(subscription_id: nil, credentials_file: DEFAULT_FILE, **_)
+          credentials_file = DEFAULT_FILE if credentials_file.nil?
```
phiggins: Idiomatic ruby would have this as `credentials_file ||= DEFAULT_FILE`, but this should get you the same result.

PR #338, opened 2018-08-17
Ensure unique_identifier returns something meaningful for service acc…

skpaterson:
…ounts.

### DIFF
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -84,8 +84,12 @@ module Train::Transports
       end
 
       def unique_identifier
-        # use auth client_id - same to retrieve for any of the clients but use IAM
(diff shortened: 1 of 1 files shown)


tomqwu:
@jquick, do you have any idea when this will be released to gem? and I assume a PR downstream to InSpec?

jquick:
Released a new train gem today, and releasing inspec with it now!

tomqwu:
@jquick indeed this fixes the issue as I was doing some test locally by building inspec from source with latest gem, here is the PR https://github.com/inspec/inspec/pull/3331 for bumping the version if it's not yet

jquick:
Gotcha, yeah it should update if you do another bundle install but we can pin up for the next release

Issue #110, opened 2016-05-24
test remote host via ipv6 address


atomic111:
I want to test a remote host via ipv6 address and i got the following output:

```
inspec shell -t ssh://pmuench@2000:111b:ffe2:1:0:0:0:2 --port 545 --key-files /home/user/.ssh/id_rsa
/usr/lib/ruby/2.3.0/uri/rfc3986_parser.rb:67:in `split': bad URI(is not URI?): ssh://pmuench@2000:111b:ffe2:1:0:0:0:2 (URI::InvalidURIError)
        from /usr/lib/ruby/2.3.0/uri/rfc3986_parser.rb:73:in `parse'
        from /usr/lib/ruby/2.3.0/uri/common.rb:227:in `parse'
        from /usr/lib/ruby/gems/2.3.0/gems/r-train-0.11.3/lib/train.rb:66:in `target_config'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/lib/inspec/backend.rb:16:in `create'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/lib/inspec/runner.rb:45:in `configure_transport'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/lib/inspec/runner.rb:29:in `initialize'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/lib/inspec/cli.rb:141:in `new'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/lib/inspec/cli.rb:141:in `shell_func'
        from /usr/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/command.rb:27:in `run'
        from /usr/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/invocation.rb:126:in `invoke_command'
        from /usr/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor.rb:359:in `dispatch'
        from /usr/lib/ruby/gems/2.3.0/gems/thor-0.19.1/lib/thor/base.rb:440:in `start'
        from /usr/lib/ruby/gems/2.3.0/gems/inspec-0.21.1/bin/inspec:9:in `<top (required)>'
        from /home/user/.gem/ruby/2.3.0/bin/inspec:22:in `load'
        from /home/user/.gem/ruby/2.3.0/bin/inspec:22:in `<main>'
```

[...]

PR #339, opened 2018-08-21
Use Addressable::URI, allowing IPv6 URLs

mgrobelin:
"Addressable is a replacement for the URI implementation that is part of Ruby's standard library. It more closely conforms to RFC 3986, RFC 3987, and RFC 6570 (level 4), additionally providing support for IRIs and URI templates.", see https://github.com/sporkmonger/addressable

Fixes https://github.com/inspec/train/issues/110

Signed-off-by: Markus Grobelin <grobi@koppzu.de>

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,8 +12,6 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.2')
   gem 'json', '< 2.0'
 end
 
-gem 'addressable', '~> 2.5'
-
 group :test do
   gem 'bundler', '~> 1.11'
   gem 'minitest', '~> 5.8'
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,8 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.2')
   gem 'json', '< 2.0'
 end
 
+gem 'addressable', '~> 2.5'
+
 group :test do
   gem 'bundler', '~> 1.11'
   gem 'minitest', '~> 5.8'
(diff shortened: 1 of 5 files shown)


mgrobelin:
I didn't  add `gem 'addressable'` to the Gemfile, as it's already dependent. Shall I?

jquick:
Thanks for this @mgrobelin ! Yeah if its part of Ruby's standard library you should not need to gem it in.

Gemfile
```diff
@@ -12,6 +12,8 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.2')
   gem 'json', '< 2.0'
 end
 
+gem 'addressable', '~> 2.5'
```
clintoncwolfe: This isn't a development need - it should only be in the gemspec.  Gemfile pulls in all gemspec requirements, so this is redundant.

train.gemspec
```diff
@@ -25,6 +25,7 @@ Gem::Specification.new do |spec|
 
   spec.required_ruby_version = '>= 2.0'
 
+  spec.add_dependency 'addressable', '~> 2.5'
```
tas50: The pin here on 2.5 is going to conflict with Chef which is pinned to 2.4 due to another gem. Does this need to be 2.5 or would 2.4 work?

mgrobelin: Hmm, addressable 2.4 is from end of 2015. Maybe I was required to bump the gem while working on a project for Ruby 2.5 support, but maybe I remember wrong. 

For sure a lot of fixes happened since 2.4. I don't know whether the change to addressable gem is even smart when not used with 2.5

tas50: We're currently waiting on Travis to release this gem so we can pull in the latest addressable to our overall stack: https://github.com/travis-ci/gh/issues/33

zenspider:
Closing in favor of #566, which I'll try to merge soon.

PR #351, opened 2018-09-19
Adds Azure Vault Client

r-fennell:
### Description
Adds functionality to auth against Azure Vault and return a Client for talking to a given Vaults APIs.

### Check List

- [x] New functionality includes tests
- [x] All tests pass
- [x] `rake lint` passes
- [x] All commits have been signed-off for the Developer Certificate of Origin.


lib/train/transports/azure.rb
```diff
@@ -59,15 +61,17 @@ def platform
         force_platform!('azure', @platform_details)
       end
 
-      def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client)
-        if cache_enabled?(:api_call)
+      def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client, use_cache = true, vault_name = nil)
```
jquick: I don't think we need the `use_cache` here. You can turn off caching easily by calling:
```
connection.disable_cache(:api_call)
connection.azure_client(...)
```

lib/train/transports/clients/azure/graph_rbac.rb
```diff
@@ -1,28 +1,32 @@
 # encoding: utf-8
+# frozen_string_literal: true
```
jquick: Since we still support ruby 2.2 please use string.freeze instead.

lib/train/transports/clients/azure/graph_rbac.rb
```diff
@@ -1,28 +1,32 @@
...
 
 # Wrapper class for ::Azure::GraphRbac::Profiles::Latest::Client allowing custom configuration,
 # for example, defining additional settings for the ::MsRestAzure::ApplicationTokenProvider.
 class GraphRbac
-  AUTH_ENDPOINT = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_endpoint_url
-  API_ENDPOINT = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_graph_resource_id
+  def initialize
+    @auth_endpoint = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_endpoint_url
+    @api_endpoint = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_graph_resource_id
+  end
```
TrevorBramble: I can't see why you want to make this class stateful. What's the benefit?

lib/train/transports/clients/azure/vault.rb
```diff
@@ -0,0 +1,36 @@
...
+
+# Wrapper class for ::Azure::KeyVault::Profiles::Latest::Mgmt::Client allowing custom configuration,
+# for example, defining additional settings for the ::MsRestAzure::ApplicationTokenProvider.
+class Vault
+  def initialize(vault_name)
+    if vault_name.nil?
+      raise ::Train::UserError, 'Vault Name cannot be nil'
+    end
+    @auth_endpoint = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_endpoint_url
+    @api_endpoint = "https://#{vault_name}#{MsRestAzure::AzureEnvironments::AzureCloud.key_vault_dns_suffix}"
```
TrevorBramble: Is this missing a separator between the interpolated values? A period?

lib/train/transports/clients/azure/vault.rb
```diff
@@ -0,0 +1,36 @@
...
+    @resource_endpoint = 'https://vault.azure.net'
+  end
```
TrevorBramble: Ah, was `GraphRbac` changed to make it match this API? If so, I don't think this should be stateful either. Both of these classes take some input and return the appropriate client, and are then discarded. No further input or other reason for change over time.

```ruby
class Vault
  AUTH_ENDPOINT     = MsRestAzure::AzureEnvironments::AzureCloud.active_directory_endpoint_url
  RESOURCE_ENDPOINT = 'https://vault.azure.net'

  def self.client(vault_name, credentials)
    raise ::Train::UserError, 'Vault Name cannot be nil' unless vault_name

    ::Azure::KeyVault::Profiles::Latest::Mgmt::Client.new(
      credentials.merge(
        credentials: ::MsRest::TokenCredentials.new(provider(credentials, settings)),
￼        base_url: api_endpoint(vault_name)
      )
    )
  end

  def self.api_endpoint(vault_name)
    "https://#{vault_name}#{MsRestAzure::AzureEnvironments::AzureCloud.key_vault_dns_suffix}"
  end

  def self.provider(credentials, settings)
    ::MsRestAzure::ApplicationTokenProvider.new(
￼       credentials[:tenant_id],
￼       credentials[:client_id],
￼       credentials[:client_secret],
￼       settings,
￼     )
  end

  def self.settings
￼     client_settings = ::MsRestAzure::ActiveDirectoryServiceSettings.get_azure_settings
￼     client_settings.authentication_endpoint = AUTH_ENDPOINT
￼     client_settings.token_audience = RESOURCE_ENDPOINT
￼     client_settings
￼   end
end
```

You could just make it a module then, of course, and replace the `self.` prefixes with entries in `module_function`, making the internal implementation methods private with `private_class_method`. But that's all a bit out of step with the conventions in this library.
```ruby
module Vault
  def client(vault_name, credentials)
  # ...

  def api_endpoint(vault_name)
  # ...

  def provider(credentials, settings)
  # ...

  def settings
  # ...

  module_function :client, :api_endpoint, :provider, :settings
  private_class_method :api_endpoint, :provider, :settings
end
```

lib/train/transports/clients/azure/graph_rbac.rb
r-fennell: You're right - it was unnecesary, i've changed it back!

lib/train/transports/clients/azure/vault.rb
r-fennell: the suffix adds on the leading `.` :)

lib/train/transports/clients/azure/vault.rb
r-fennell: Thanks for this! I went for something very similar - will ping you about it.

lib/train/transports/clients/azure/vault.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  private_class_method def self.api_endpoint(vault_name)
+    "https://#{vault_name}#{MsRestAzure::AzureEnvironments::AzureCloud.key_vault_dns_suffix}"
+  end
+
+  private_class_method def self.settings
+    client_settings = MsRestAzure::ActiveDirectoryServiceSettings.get_azure_settings
+    client_settings.authentication_endpoint = AUTH_ENDPOINT
+    client_settings.token_audience = RESOURCE_ENDPOINT
+    client_settings
```
TrevorBramble: The assignment, modification, and finally return of the local variable is an anti-pattern, and I'm sad to see that actually comes from [their library](https://github.com/Azure/azure-sdk-for-ruby/blob/f1160ec0e3193c0e1a1066776b415ee009f655a7/runtime/ms_rest_azure/lib/ms_rest_azure/active_directory_service_settings.rb#L57-L60).

How about we just give `get_azure_settings` something that satisfies the interface instead of doing those (re-)assignments on the returned value?
```ruby
def self.settings
  MsRestAzure::ActiveDirectoryServiceSettings.get_azure_settings(
    Struct.new(:authentication_endpoint, :token_audience)
          .new(AUTH_ENDPOINT, RESOURCE_ENDPOINT)
  )
end
```

lib/train/transports/azure.rb
```diff
@@ -59,15 +61,17 @@ def platform
         force_platform!('azure', @platform_details)
       end
 
-      def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client)
+      def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client, vault_name = nil)
```
jquick: I would prefer valut_name to be a hash here so its flex-able if we want to add other options for clients:
```
azure_client(Azure::KeyVault::Profiles::Latest::Mgmt::Client, vault_name: vault1, valut_id: 1234)

def azure_client(klass = ::Azure::Resources::Profiles::Latest::Mgmt::Client, opts = {})
...
  client = Vault.client(opts[:vault_name], @credentials)
end
```

lib/train/transports/azure.rb
```diff
@@ -59,15 +61,17 @@ def platform
...
           return @cache[:api_call][klass.to_s.to_sym] unless @cache[:api_call][klass.to_s.to_sym].nil?
         end
 
         if klass == ::Azure::Resources::Profiles::Latest::Mgmt::Client
           @credentials[:base_url] = MsRestAzure::AzureEnvironments::AzureCloud.resource_manager_endpoint_url
         elsif klass == ::Azure::GraphRbac::Profiles::Latest::Client
-          client =  GraphRbac.client(@credentials)
+          client = GraphRbac.client(@credentials)
+        elsif klass == ::Azure::KeyVault::Profiles::Latest::Mgmt::Client
+          client = Vault.client(vault_name, @credentials)
```
jquick: Does Azure Vault support MSI connection? If not we may want to raise somewhere saying its not supported.

jquick:
Closing and reopening to trigger tests again.

jquick:
Travis builds passing https://travis-ci.com/inspec/train/builds/87510782. This is good to go.

PR #361, opened 2018-10-03
Fix detection of 'SLES Expanded Support' of Red Hat

jabofh:
* SLES tags Expanded Support versions in /etc/os-release with comments
* This broke platform detection

### DIFF
--- a/lib/train/platforms/detect/helpers/os_linux.rb
+++ b/lib/train/platforms/detect/helpers/os_linux.rb
@@ -37,7 +37,7 @@ module Train::Platforms::Detect::Helpers
 
       raw.lines.each_with_object({}) do |line, memo|
         line.strip!
-        next if line.start_with?("#")
+        next if line.start_with?('#')
         next if line.empty?
         key, value = line.split('=', 2)
         memo[key] = value.gsub(/\A"|"\Z/, '') unless value.empty?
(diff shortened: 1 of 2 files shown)


chris-rock:
Welcome @jabofh to the InSpec community. Great to see this fix. Can you help us writing a test for the platform detection to ensure this will not happen again?

Tests are located here https://github.com/inspec/train/blob/79badbfc932a51c29c5f0d4dd19f64074d554a70/test/unit/platforms/os_detect_test.rb#L35-L68

chris-rock:
Please sign-off your commit as described in https://github.com/inspec/train/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

jabofh:
Have done so, thanks...

jabofh:
I have added a test for SLES ESR RedHat; it seems fine.

lib/train/platforms/detect/helpers/os_linux.rb
```diff
@@ -37,6 +37,7 @@ def parse_os_release_info(raw)
 
       raw.lines.each_with_object({}) do |line, memo|
         line.strip!
+        next if line.start_with?("#")
```
clintoncwolfe: This line here will trip the linter. Change the `"` to `'` and it will pass.  If the linter fails, the CI fails, and if the CI fails, we can't accept it.

I understand if you son't want to change a quote sign after so long- I'll flag this as an adoption candidate.

jabofh: I have updated the code... Took a while to remember what is was about, since I have been running my fork in the meantime...

PR #364, opened 2018-10-05
Small style/spelling changes for Train example plugin

jerryaldrichiii:


examples/plugins/train-local-rot13/lib/train-local-rot13/connection.rb
```diff
@@ -52,11 +52,6 @@ def initialize(options)
         # credentials, now is a good time.
       end
 
-      # If you are writing a local-style connection, implement this to return true.
```
clintoncwolfe: Please explain this in a comment in lib/train/plugins/base_connection.rb . You learned a lot of tribal knowledge from Jared about why this isn't needed - record it in the superclass.

jerryaldrichiii: Oooooh, great call! I'll make that happen.

jerryaldrichiii: Better yet, I think I can just remove it all together. I'll submit a PR.

jquick:
re-doing tests

clintoncwolfe:
Ignoring unrelated unit test failures due to #393

PR #365, opened 2018-10-10
Remove `#local?`

jerryaldrichiii:
It was used only once and provides more confusion than benefit.

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -17,9 +17,17 @@ module Train::Platforms::Detect::Specifications
 
       plat.family('windows').in_family('os')
           .detect {
-            if winrm? || (@backend.local? && ruby_host_os(/mswin|mingw32|windows/))
-              true
+            # Can't return from a `proc` thus the `is_windows` shenanigans
+            is_windows = false
+            is_windows = true if winrm?
+
+            if defined?(Train::Transports::Local::Connection)
+              if @backend.class == Train::Transports::Local::Connection
+                is_windows = true
+              end
             end
+
+            is_windows
           }
       # windows platform
       plat.name('windows').in_family('windows')
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -21,10 +21,8 @@ module Train::Platforms::Detect::Specifications
             is_windows = false
             is_windows = true if winrm?
 
-            if defined?(Train::Transports::Local::Connection)
-              if @backend.class == Train::Transports::Local::Connection
-                is_windows = true
-              end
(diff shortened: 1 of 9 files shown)


TrevorBramble:
It's a little early to say, but we're already talking about querying the HTTP service locally and having to SSH before querying for remote hosts. Is there some other mechanism by which this would be done?

jerryaldrichiii:
@TrevorBramble this shouldn't affect that. This only removes the `local?` method on the transport. It was only used in one place as a workaround to a detect issue on Windows. This removes that and handles the workaround in another way.

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -17,9 +17,17 @@ def self.load
...
-            if winrm? || (@backend.local? && ruby_host_os(/mswin|mingw32|windows/))
-              true
+            # Can't return from a `proc` thus the `is_windows` shenanigans
+            is_windows = false
+            is_windows = true if winrm?
+
+            if defined?(Train::Transports::Local::Connection)
+              if @backend.class == Train::Transports::Local::Connection
+                is_windows = true
+              end
```
jquick: This could also be `if winrm? || (@backend.class.to_s == 'Train::Transports::Local::Connection' && ruby_host_os(/mswin|mingw32|windows/))`

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -17,9 +17,17 @@ def self.load
...
       plat.family('windows').in_family('os')
           .detect {
-            if winrm? || (@backend.local? && ruby_host_os(/mswin|mingw32|windows/))
-              true
+            # Can't return from a `proc` thus the `is_windows` shenanigans
+            is_windows = false
+            is_windows = true if winrm?
+
+            if defined?(Train::Transports::Local::Connection)
+              if @backend.class == Train::Transports::Local::Connection
```
jquick: You still need to check for `ruby_host_os(/mswin|mingw32|windows/)` here

lib/train/platforms/detect/specifications/os.rb
jerryaldrichiii: I thought that, but you have to check if it is defined as well otherwise you get an error.

jquick: the to_s kinda gets around that but im fine either way

lib/train/platforms/detect/specifications/os.rb
jerryaldrichiii: Great catch will update.

lib/train/platforms/detect/specifications/os.rb
jerryaldrichiii: Ah! I like that more. I'll do that.

PR #367, opened 2018-10-12
Add method for `@platform[:uuid_command]` to prevent setting it to `'unknown'`

jerryaldrichiii:
Currently, if you were to create a connection that did not provide `#unique_identifier` then `@platform[:uuid_command]` would return `'unknown'` due to backwards compatibility. See: https://github.com/inspec/train/blob/master/lib/train/platforms/platform.rb#L58

This makes a an accesser for `uuid_command` so that  `@platform[:uuid_command]` can return `nil`.

Consequently, we also need to raise something other than `StandardError` so it can be caught upstream (e.g. InSpec)

### DIFF
--- a/lib/train/platforms/detect/scanner.rb
+++ b/lib/train/platforms/detect/scanner.rb
@@ -6,12 +6,10 @@ module Train::Platforms::Detect
   class Scanner
     include Train::Platforms::Detect::Helpers::OSCommon
 
-    attr_accessor :uuid_command # Used by mock transport during testing
     def initialize(backend)
       @backend = backend
       @platform = {}
       @family_hierarchy = []
-      @uuid_command = nil
 
       # detect cache variables
       @files = {}
--- a/lib/train/platforms/detect/scanner.rb
+++ b/lib/train/platforms/detect/scanner.rb
@@ -6,10 +6,12 @@ module Train::Platforms::Detect
   class Scanner
     include Train::Platforms::Detect::Helpers::OSCommon
 
+    attr_accessor :uuid_command # Used by mock transport during testing
     def initialize(backend)
       @backend = backend
       @platform = {}
       @family_hierarchy = []
+      @uuid_command = nil
 
       # detect cache variables
       @files = {}
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -461,7 +461,7 @@ module Train::Platforms::Detect::Specifications
       plat.name('mac_os_x').title('macOS X').in_family('darwin')
           .detect {
             cmd = unix_file_contents('/System/Library/CoreServices/SystemVersion.plist')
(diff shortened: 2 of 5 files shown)


lib/train/platforms/detect/uuid.rb
```diff
@@ -22,12 +22,12 @@ def find_or_create_uuid
...
       else
-        if @platform[:uuid_command]
-          result = @backend.run_command(@platform[:uuid_command])
+        if @platform.uuid_command
+          result = @backend.run_command(@platform.uuid_command)
           return uuid_from_string(result.stdout.chomp) if result.exit_status.zero? && !result.stdout.empty?
         end
 
-        raise 'Could not find platform uuid! Please set a uuid_command for your platform.'
+        raise Train::TransportError, 'Could not find platform uuid! Please set a uuid_command for your platform.'
```
jerryaldrichiii: I think we should remove this raise all together...but I'll put it up for discussion.

Currently this shows as a warning on an InSpec run.

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -461,7 +461,7 @@ def self.load
       plat.name('mac_os_x').title('macOS X').in_family('darwin')
           .detect {
             cmd = unix_file_contents('/System/Library/CoreServices/SystemVersion.plist')
-            @platform[:uuid_command] = "system_profiler SPHardwareDataType | awk '/UUID/ { print $3; }'"
+            self.uuid_command = "system_profiler SPHardwareDataType | awk '/UUID/ { print $3; }'"
```
jquick: should this be `@platform.uuid_command`? I think self works but it is a bit hard to figure out who is self when you'r passing around these procs.

lib/train/platforms/detect/scanner.rb
```diff
@@ -6,10 +6,12 @@ module Train::Platforms::Detect
   class Scanner
     include Train::Platforms::Detect::Helpers::OSCommon
 
+    attr_accessor :uuid_command # Used by mock transport during testing
```
jquick: I think this should be a attr_accessor on the platform not the scanner.

lib/train/platforms/detect/uuid.rb
jquick: 3.0 removes any uuid pulling so it will not anymore

jerryaldrichiii:
In the detect blocks `@platform` is a hash so I cannot update those to be `@platform.uuid_command`. Now that we've added this method though it will make a call to `@platform.uuid_command` anyway due to: https://github.com/inspec/train/blob/master/lib/train/platforms/platform.rb#L58

Let me know if I'm misunderstanding. This was giving me quite a bit of trouble.

clintoncwolfe:
As this PR appears to touch functionality that has been deprecated, and has not been updated in a while, I'm going to close it.  If I'm mistaken, please open an issue describing the problem in detail, then re-open this or make a new PR to fix the problem.

PR #371, opened 2018-10-22
Adds cached_client method in BaseConnection

dmccown:
Adds a helper method in base connection so you no longer have to
duplicate caching logic in any plugins you may create. Example:

```Ruby
def demo_client
  cache_client(:api_call, :demo_connection) do
    DemoClient.new(options)
  end
end
```

This will automatically check if the cache is enabled and return the the
cached object if it exists. If it doesn't exist then the cache will be
hydrated.

### DIFF
--- a/lib/train/plugins/base_connection.rb
+++ b/lib/train/plugins/base_connection.rb
@@ -37,6 +37,25 @@ class Train::Plugins::Transport
       end
     end
 
+    # Returns cached client if caching enabled. Otherwise returns whatever is
+    # given in the block.
+    #
+    # @example
+    #
+    #   def demo_client
+    #     cached_client(:api_call, :demo_connection) do
+    #       DemoClient.new(args)
+    #     end
+    #   end
+    #
+    # @param [symbol] type one of [:api_call, :file, :command]
+    # @param [symbol] key for your cached connection
+    def cached_client(type, key)
+      return yield unless cache_enabled?(type)
+
+      @cache[type][key] ||= yield
+    end
+
     def cache_enabled?(type)
       @cache_enabled[type.to_sym]
     end
--- a/test/unit/plugins/connection_test.rb
+++ b/test/unit/plugins/connection_test.rb
@@ -78,6 +78,54 @@ describe 'v1 Connection Plugin' do
          .method(:logger).call.must_equal(l)
     end
 
+    describe 'cached_client helper' do
+      class DemoConnection < Train::Plugins::Transport::BaseConnection
+        def initialize(options = {})
+          super(options)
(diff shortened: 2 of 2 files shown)


lib/train/plugins/base_connection.rb
```diff
@@ -37,6 +37,25 @@ def initialize(options = nil)
...
+    #       DemoClient.new(args)
+    #     end
+    #   end
+    #
+    # @param [symbol] type one of [:api_call, :file, :command]
+    # @param [symbol] key for your cached connection
+    def cached_client(type, key)
+      return yield unless cache_enabled?(type)
+
+      @cache[type][key] ||= yield
```
TrevorBramble: `@cache.dig(type, key) ||= yield`?

dmccown: I wasn't sure if you could do assignment when using dig.

test/unit/plugins/connection_test.rb
```diff
@@ -78,6 +78,54 @@
...
+        client2 = conn.demo_client
+
+        client1.wont_be_same_as client2
+      end
+
+      it 'returns a new connection when cache enabled and not hydrated' do
+        conn = DemoConnection.new
+        conn.enable_cache(:api_call)
+
+        client1 = conn.demo_client
```
TrevorBramble: It wouldn't be useful in the other tests, because they don't look anything like real world code, but this one made me wonder if would just return `self` from `enable_cache()` and `disable_cache()`. I quick search shows that the implicit return of a boolean is never treated as significant.

```ruby
def enable_cache(type)
  fail Train::UnknownCacheType, "#{type} is not a valid cache type" unless @cache_enabled.keys.include?(type.to_sym)

  @cache_enabled[type.to_sym] = true

  self
end
```
```ruby
it 'returns a new connection when cache enabled and not hydrated' do
  conn = DemoConnection.new.enable_cache(:api_call)
￼
  client1 = conn.demo_client
  client1.must_be_instance_of DemoConnection::DemoClient
end

# or even...

it 'returns a new connection when cache enabled and not hydrated' do
  client1 = DemoConnection.new.enable_cache(:api_call).demo_client
￼
  client1.must_be_instance_of DemoConnection::DemoClient
end
```

lib/train/plugins/base_connection.rb
dmccown: That doesn't work. You aren't allowed to do assignment from the return of `dig`. 

```
train/lib/train/plugins/base_connection.rb:56: syntax error, unexpected tOP_ASGN,
expecting keyword_end (SyntaxError)
      @cache.dig(type, key) ||= yield
```

PR #378, opened 2018-11-06
Add Google API application info

nathenharvey:
This will help measure requests to the Google API that come from train.

Signed-off-by: Nathen Harvey <nathenharvey@google.com>

### DIFF
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -88,6 +88,8 @@ module Train::Transports
         scopes = ['https://www.googleapis.com/auth/cloud-platform',
                   'https://www.googleapis.com/auth/compute']
         authorization = Google::Auth.get_application_default(scopes)
+        Google::Apis::ClientOptions.default.application_name = 'chef-inspec-train'
+        Google::Apis::ClientOptions.default.application_version = Train::VERSION
         Google::Apis::RequestOptions.default.authorization = authorization
       end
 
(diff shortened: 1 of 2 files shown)


lib/train/transports/gcp.rb
```diff
@@ -88,6 +88,8 @@ def connect
         scopes = ['https://www.googleapis.com/auth/cloud-platform',
                   'https://www.googleapis.com/auth/compute']
         authorization = Google::Auth.get_application_default(scopes)
+        Google::Apis::ClientOptions.default.application_name = 'chef-train'
```
chris-rock: Should we use inspec here, since users may not know train. Alternately I propose to use `chef-inspec-train`

nathenharvey: @chris-rock I did struggle with exactly what name should be given.  Is train used by anything other than InSpec at the moment?  Will it be in the future? 

It might be better if this was something that InSpec, or other libraries, could pass to train when making connections.  

Either way, I would be happy with this:

```suggestion
        Google::Apis::ClientOptions.default.application_name = 'chef-inspec-train'
```

chris-rock: I am also not 100% sure. other Chef products use train, but not the cloud plugins. I recommend we go with `chef-inspec-train` for now and make it customizable via a plugin option for the using application going forward

nathenharvey: pushed an update with this new `application_name`

nathenharvey:
squashed the commits so it's a single commit merge when you're ready @chris-rock

PR #380, opened 2018-11-15
Added a new matcher for amazon linux 2

artyomtkachenko:
Th target of this PR is to address a problem which appeared after AWS updated naming schema for Amazon Linux 2.

More details can be found [here](https://github.com/inspec/inspec/issues/3614)

### DIFF
--- a/lib/train/platforms/detect/helpers/os_linux.rb
+++ b/lib/train/platforms/detect/helpers/os_linux.rb
@@ -12,6 +12,8 @@ module Train::Platforms::Detect::Helpers
         /((\d+) \(Rawhide\))/i.match(conf)[1].downcase
       when /Amazon Linux AMI/i
         /release ([\d\.]+)/.match(conf)[1]
+      when /amazon linux release/i
+        /Linux release ((\d+|\.)+)/i.match(conf)[1]
       when /derived from .*linux|amazon/i
         /Linux ((\d+|\.)+)/i.match(conf)[1]
       else
--- a/test/unit/platforms/detect/os_linux_test.rb
+++ b/test/unit/platforms/detect/os_linux_test.rb
@@ -28,6 +28,14 @@ describe 'os_linux' do
     it 'normal linux' do
       detector.redhatish_version('derived from Ubuntu Linux 11').must_equal('11')
     end
+
+    it 'amazon linux 2 new release naming schema' do
+      detector.redhatish_version('Amazon Linux release 2 (Karoo)').must_equal('2')
+    end
+
+    it 'amazon linux 2 old release naming schema' do
+      detector.redhatish_version('Amazon Linux 2').must_equal('2')
+    end
   end
 
   describe 'lsb parse' do


chris-rock:
@artyomtkachenko Can you please sign your commit?

chris-rock:
Can you also share where and how you experienced it? While I found references on the next, I could not reproduce the issue with latest amazon 2 docker image.

Related to https://github.com/ansible/ansible/issues/48823

radhus:
I experience this and can confirm that this fix solves the problem. I just run an EC2 instance with the latest Amazon Linux 2 AMI, fully updated, then run the latest inspec with some tests, and this started crashing since a week or so.

chris-rock:
@artyomtkachenko Is there any way we can help you? We would love to get your PR in

jerryaldrichiii:
Just realized I solved this in https://github.com/inspec/train/pull/385 ...sorry I didn't see this PR.

@artyomtkachenko in order to get you credit for this (which I really want to do :smile:) you can do the following:
  - Update this PR with the RegEx from https://github.com/inspec/train/pull/385/files (yours works due to the `else`, but I think mine is a bit cleaner)
  - Do a `git commit amend -s` and `git push origin master -f` (DCO sign off) (I'll pull your commit over and into mine)
  - Leave a comment here that you don't mind being in the Git history (I'll still give you a shout out in the message/title field)

artyomtkachenko:
Please accept my apologies Guys for a delay in a reply (all Github messages went into a SPAM folder for some reason ) and for a missing DCO.  I updated my PR with the cleaner solution from @jerryaldrichiii . 
Thank you very much again 
👍

jerryaldrichiii:
@artyomtkachenko the RegEx is still a bit off. `AMI` is not always present.

Can you change to have that method look like:

```ruby
    def redhatish_version(conf)
      case conf
      when /rawhide/i
        /((\d+) \(Rawhide\))/i.match(conf)[1].downcase
      when /Amazon Linux/i
        /([\d\.]+)/.match(conf)[1]
      when /derived from .*linux|amazon/i
        /Linux ((\d+|\.)+)/i.match(conf)[1]
      else
        /release ([\d\.]+)/.match(conf)[1]
      end
    end
```

artyomtkachenko:
> @artyomtkachenko the RegEx is still a bit off. `AMI` is not always present.
> 
> Can you change to have that method look like:
> 
> ```ruby
>     def redhatish_version(conf)
>       case conf
>       when /rawhide/i
>         /((\d+) \(Rawhide\))/i.match(conf)[1].downcase
>       when /Amazon Linux/i
>         /([\d\.]+)/.match(conf)[1]
>       when /derived from .*linux|amazon/i
>         /Linux ((\d+|\.)+)/i.match(conf)[1]
>       else
>         /release ([\d\.]+)/.match(conf)[1]
>       end
>     end
> ```

Sorry, my mstake. Forgot to run `git add` ....

PR #381, opened 2018-11-21
Pass logger to Cisco IOS transport

btm:
Inherit the logger (and log level) from upstream rather than presuming
we'll make a new one.

Signed-off-by: Bryan McLellan <btm@loftninjas.org>


btm:
Travis failures look related to GCP.

jquick:
Validated tests are working locally. Woking on a fix. Approving all tests pass for now.

PR #383, opened 2018-11-21
Fixes enable password and catches failure

### LINKED ISSUE #382: Cisco enable password silently failing
### Description

enable isn't succeeding

### Train and Platform Version

train 1.5.6
Cisco IOS Software, C3560 Software (C3560-IPSERVICESK9-M), Version 15.0(2)SE4, RELEASE SOFTWARE (fc1)

### Replication Case

run a command via train that requires enable, e.g. `show running-config`

### Possible Solutions

sending `"enable\n#{@enable_password}"` instead of `"enable\r\n#{@enable_password}"` seems to work.

### Stacktrace

```
#<struct Train::Extras::CommandResult stdout="Current privilege level is 1", stderr="", exit_status=0>
#<struct Train::Extras::CommandResult stdout="enable\r\nPassword: \r\n% Access denied\r\n\r\noffice-cisco>cisco\r\nTranslating \"cisco\"...domain server (8.8.8.8)\r\n% Unknown command or computer name, or unable to find computer address", stderr="", exit_status=0>
#<struct Train::Extras::CommandResult stdout="", stderr="", exit_status=0>
#<struct Train::Extras::CommandResult stdout="", stderr="                   ^\r\n% Invalid input detected at '^' marker.", exit_status=1>
```

btm:
On my switch, '\r\n' is being interpreted as two enter key presses, so
the enable password is being sent to the shell prompt on the switch, not
to the password prompt.

I'm not sure if it is different behavior on some routers or if this has
been broken for a while, given that inspec may not need enable. But
'show running-config' does require enable privileges.

Fixes #382

Signed-off-by: Bryan McLellan <btm@loftninjas.org>

### DIFF
--- a/lib/train/transports/cisco_ios_connection.rb
+++ b/lib/train/transports/cisco_ios_connection.rb
@@ -46,7 +46,8 @@ class Train::Transports::SSH
         # This verifies we are not in privileged exec mode before running the
         # enable command. Otherwise, the password will be in history.
         if run_command_via_connection('show privilege').stdout.split[-1] != '15'
-          run_command_via_connection("enable\n#{@enable_password}")
+          # Extra newlines to get back to prompt if incorrect password is used
+          run_command_via_connection("enable\n#{@enable_password}\n\n\n")
         end
       end
 
@@ -65,7 +66,9 @@ class Train::Transports::SSH
 
       logger.debug('[SSH] waiting for prompt')
       until @buf =~ @prompt
-        raise BadEnablePassword if @buf =~ /Bad secrets/ || @buf =~ /Access denied/
+        if @buf =~ /Bad (secrets|password)|Access denied/
+          raise BadEnablePassword
+        end
         session.connection.process(0)
       end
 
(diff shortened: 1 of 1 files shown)


btm:
@jerryaldrichiii thoughts on this?

jerryaldrichiii:
Interesting, I've only tested this on IOS devices in GNS3. What you're doing seems correct though. I'll give it a test on my lab and see if it doesn't break anything.

Thanks for this!

clintoncwolfe:
Hi! As this fix has been sitting for a while, and requires access to specialized testing environments, the InSpec team is not going to be able to accept this as-is.  If you have time to update it, please do re-open, otherwise I'm going to close this for now.

btm:
appveyor failure looks like it's on master too. This should be ready. This method lacks any unit test framework and since it's over a transport it doesn't look easy to add right now.

lib/train/transports/cisco_ios_connection.rb
```diff
@@ -65,7 +66,9 @@ def run_command_via_connection(cmd, &_data_handler)
 
       logger.debug('[SSH] waiting for prompt')
       until @buf =~ @prompt
-        raise BadEnablePassword if @buf =~ /Bad secrets/
+        if @buf =~ /Bad [secrets|password]/ || @buf =~ /Access denied/
```
zenspider: This regexp looks wrong to me. Not sure if you want an alternation or to escape the square brackets but I'm guessing the former:

```ruby
        if @buf =~ /Bad (secrets|password)/ || @buf =~ /Access denied/
```

and then at that point, they might as well merge:

```ruby
        if @buf =~ /Bad (secrets|password)|Access denied/
```

Issue #391, opened 2018-12-12
Expose additional winrm options


frezbo:
### Description

On FIPS enabled clients the WinRM transport can only use `:ssl` **transport** and `basic_auth`. NTLM auth uses `MD4` is is not at all feasible.

 ### Train and Platform Version

All verisons

### Replication Case

Ruby just barks when using `MD5` in FIPS mode. `MD4` is not at all available. 

The only working config on FIPS machine:

```
require 'winrm'

opts =
    {
    transport: :ssl,
    endpoint: 'https://<some server>:5986/wsman',
    user: 'Administrator',
    password: '<redacted>,
    no_ssl_peer_verification: true,
    basic_auth_only: true
    }

conn = WinRM::Connection.new(opts)

shell = conn.shell(:powershell)
puts shell.run('ls').stdout
shell.close
```

PR #392, opened 2018-12-12
Expose additional winrm options

frezbo:
Signed-off-by: Noel Georgi <git@frezbo.com>

Fixes: #391

### DIFF
--- a/lib/train/transports/winrm.rb
+++ b/lib/train/transports/winrm.rb
@@ -40,14 +40,19 @@ module Train::Transports
 
     require 'train/transports/winrm_connection'
 
+    # ref: https://github.com/WinRb/Winrm#transports
+    # kerberos is skipped here as the options are not exposed in train
+    # basic ssl options is only supported
+    SUPPORTED_WINRM_TRANSPORTS = %i[negotiate ssl plaintext].freeze
+
     # common target configuration
     option :host, required: true
     option :port
     option :user, default: 'administrator', required: true
     option :password, nil
-    option :transport, default: :negotiate
-    option :disable_sspi, default: false
-    option :basic_auth_only, default: false
+    option :winrm_transport, default: :negotiate
+    option :winrm_disable_sspi, default: false
+    option :winrm_basic_auth_only, default: false
     option :path, default: '/wsman'
     option :ssl, default: false
     option :self_signed, default: false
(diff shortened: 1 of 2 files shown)


frezbo:
rerunning to fix the GCP fluke.

frezbo:
The failure is related to GCP test suite.

lib/train/transports/winrm.rb
```diff
@@ -45,6 +45,9 @@ class WinRM < Train.plugin(1)
     option :port
     option :user, default: 'administrator', required: true
     option :password, nil
+    option :transport, default: :negotiate
```
clintoncwolfe: For now, let's add a --winrm- prefix here, to match what InSpec will be sending.  I think in the future we may be able to something like having InSpec strip the prefix from options (or all options would have the prefix - point being it should be consistent)

lib/train/transports/winrm.rb
```diff
@@ -100,9 +103,9 @@ def validate_options(opts)
     def connection_options(opts)
       {
         logger:                   logger,
-        transport:                :negotiate,
-        disable_sspi:             false,
-        basic_auth_only:          false,
+        transport:                opts[:transport].intern,
```
clintoncwolfe: It's equivalent, but more readable, to say `.to_sym` here.

lib/train/transports/winrm.rb
clintoncwolfe: It would be great if you could add a test file in `test/windows/` to exercise the new options (those tests are run by `rake test:windows`, see Rakefile for details).  If that would be overly difficult to do, I'm OK with omitting that for now - but please open a new issue on Train mentioning the need to add testing for the new winrm options, so it doesn't get lost.

lib/train/transports/winrm.rb
clintoncwolfe: Actually, come to think of it, we might want to add some validation here; I'm not sure what valid values are for the transport flag.

lib/train/transports/winrm.rb
frezbo: resolved.

lib/train/transports/winrm.rb
frezbo: resolved.

lib/train/transports/winrm.rb
frezbo: resolved.

frezbo:
@clintoncwolfe I have updated the PR as per the review. The tests are failing due to `gcp` UT's failing. (It is also faky on the local machine, would be good to fix that). I have omitted `kerberos` transport. The `ssl` transport also exposes some additional options, but it would go out of scope of this MVP. I think it would be good to track that later. I have also added a baseline UT's for the winrm transport.

frezbo:
Re-trigger tests

clintoncwolfe:
Disregarding unrelated failing unit test per #393

PR #394, opened 2019-01-07
Prepare train for using credential sets

clintoncwolfe:
Train has historically either used explicit options passed in a hash, or parsed a --target option as a URL. Both are a bit unwieldy, especially the URI method. 

This PR makes a few changes to make it easier to work with credentials passed in as a hash.

### DIFF
--- a/lib/train.rb
+++ b/lib/train.rb
@@ -61,42 +61,8 @@ module Train
     raise ex
   end
 
-  # Resolve target configuration, using information in the config object.
-  # in URI-scheme. respective fields and merge with existing configuration.
-  # e.g. ssh://bob@remote  =>  backend: ssh, user: bob, host: remote
-  def self.unpack_target_creds(config = nil) # rubocop:disable Metrics/AbcSize
-    unless config.respond_to?(:target_uri)
-      return target_config(config)
-    end
-
-    # A set of connecting information that will make sense to
-    # the selected transports. Different transports expect different
-    # keys here.
-    credentials = { target: config.target_uri }
-    return credentials if config.target_uri.empty?
-
-    # Look for transport://credset format
-    %r{^(?<transport_name>[a-z_\-0-9]+)://(?<cred_set_name>[a-z_\-0-9]*)$} =~ config.target_uri
-
-    unless transport_name
-      # We must need to parse from a more complex URI
-      return credentials.merge(unpack_target_from_uri(config.target_uri)) # TODO: dubious
-    end
-
-    transport_name = transport_name.to_sym
-    credentials[:backend] = transport_name
-
-    transport_class = load_transport(transport_name)
-    credentials.merge(config.fetch_credentials(
-      transport_name,
-      cred_set_name.to_sym,
-      option_defaults: transport_class.credential_option_defaults,
-    ))
-    # TODO: we could ask the transport class to validate / preflight the credset
-    #  (Inspec::Backend#create calls Train.validate_backend)
-  end
-
   # Legacy code to unpack a series of items from an incoming Hash
+  # Inspec::Config.unpack_train_credentials now handles this in most cases
   def self.target_config(config) # TODO: deprecate
     conf = config.dup
     # Symbolize keys
@@ -108,10 +74,10 @@ module Train
 
     group_keys_and_keyfiles(conf) # TODO: move logic into SSH plugin
     return conf if conf[:target].to_s.empty?
-    return (unpack_target_from_uri(conf[:target], conf)).merge(conf)
+    unpack_target_from_uri(conf[:target], conf).merge(conf)
   end
 
-  def self.unpack_target_from_uri(uri_string, opts = {})
+  def self.unpack_target_from_uri(uri_string, opts = {}) # rubocop: disable Metrics/AbcSize
     creds = {}
     return creds if uri_string.empty?
 
@@ -133,6 +99,8 @@ module Train
 
     # ensure path is nil, if its empty; e.g. required to reset defaults for winrm # TODO: move logic into winrm plugin
     creds[:path] = nil if !creds[:path].nil? && creds[:path].to_s.empty?
+
+    creds.compact!
 
     # return the updated config
     creds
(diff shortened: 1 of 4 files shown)


lib/train.rb
```diff
@@ -61,53 +61,52 @@ def self.load_transport(transport_name)
...
-    conf = symbolize_keys(conf)
-
-    group_keys_and_keyfiles(conf)
+  # Legacy code to unpack a series of items from an incoming Hash
+  # Inspec::Config.unpack_train_credentials now handles this in most cases that InSpec needs
+  # If you need to unpack a URI, use unpack_target_from_uri
+  def self.target_config(config = nil)
+    conf = config.dup
+    # Symbolize keys
+    conf.keys.each do |key|
```
tas50:
```suggestion
    conf.each_key do |key|
```

lib/train.rb
```diff
@@ -61,53 +61,52 @@ def self.load_transport(transport_name)
...
-  # e.g. ssh://bob@remote  =>  backend: ssh, user: bob, host: remote
-  def self.target_config(config = nil) # rubocop:disable Metrics/AbcSize
-    conf = config.nil? ? {} : config.dup
-    conf = symbolize_keys(conf)
-
-    group_keys_and_keyfiles(conf)
+  # Legacy code to unpack a series of items from an incoming Hash
+  # Inspec::Config.unpack_train_credentials now handles this in most cases that InSpec needs
+  # If you need to unpack a URI, use unpack_target_from_uri
+  def self.target_config(config = nil)
```
clintoncwolfe: Consider adding deprecation warning

lib/train.rb
```diff
@@ -127,34 +126,38 @@ def self.parse_uri(string)
...
-    return default if conf.nil?
-    res = conf[:backend]
+  # Examine the given credential information, and if all is well,
+  # return the transport name.
+  def self.validate_backend(credentials, default_transport_name = 'local')
+    return default_transport_name if credentials.nil?
+    result = credentials[:backend]
 
-    if (res.nil? || res == 'localhost') && conf[:sudo]
+    if (result == 'local' || result == 'localhost') && credentials[:sudo]
```
clintoncwolfe: Verify that nil is no longer possible

clintoncwolfe: Consider removing localhost if that is incoherent

lib/train.rb
```diff
@@ -61,53 +61,52 @@ def self.load_transport(transport_name)
...
+        conf[key.to_sym] = conf.delete(key)
+      end
+    end
 
+    group_keys_and_keyfiles(conf) # TODO: move logic into SSH plugin
     return conf if conf[:target].to_s.empty?
+    unpack_target_from_uri(conf[:target], conf).merge(conf)
+  end
+
+  def self.unpack_target_from_uri(uri_string, opts = {}) # rubocop: disable Metrics/AbcSize
```
tas50: Can we get a comment here briefly explaining the goal of this method and why folks should be using it over the legacy method. Otherwise rubydocs won't be very useful for this.

lib/train.rb
```diff
@@ -127,34 +126,38 @@ def self.parse_uri(string)
...
   private_class_method :parse_uri
 
-  def self.validate_backend(conf, default = :local)
-    return default if conf.nil?
-    res = conf[:backend]
+  # Examine the given credential information, and if all is well,
+  # return the transport name.
+  def self.validate_backend(credentials, default_transport_name = 'local')
+    return default_transport_name if credentials.nil?
+    result = credentials[:backend]
```
clintoncwolfe: Consider renaming to a more meaningful var name - backend?

test/unit/train_test.rb
```diff
@@ -2,6 +2,7 @@
 #
 # Author:: Dominik Richter (<dominik.richter@gmail.com>)
 require_relative 'helper'
+require 'byebug'
```
clintoncwolfe: Move to helper

test/unit/train_test.rb
```diff
@@ -228,7 +229,7 @@
     end
 
     it 'returns the local backend if nothing was provided' do
-      Train.validate_backend({}).must_equal :local
+      Train.validate_backend({}).must_equal 'local'
```
clintoncwolfe: If all backend driver names are strings, how did this ever pass as symbols?

lib/train.rb
```diff
@@ -127,34 +126,38 @@ def self.parse_uri(string)
...
-    return res if !res.nil?
+    return result if !result.nil?
 
-    if !conf[:target].nil?
+    if !credentials[:target].nil?
+      # We should not get here, because if target_uri unpacking was successful,
+      # it would have set credentials[:backend]
       fail Train::UserError, 'Cannot determine backend from target '\
-           "configuration #{conf[:target].inspect}. Valid example: ssh://192.168.0.1."
+           "configuration #{credentials[:target]}. Valid example: ssh://192.168.0.1."
```
jerryaldrichiii:
```suggestion
           "configuration #{credentials[:target]}. Valid example: ssh://192.168.0.1"
```

lib/train.rb
```diff
@@ -61,53 +61,52 @@ def self.load_transport(transport_name)
...
 
-    # ensure path is nil, if its empty; e.g. required to reset defaults for winrm
-    conf[:path] = nil if !conf[:path].nil? && conf[:path].to_s.empty?
+    # ensure path is nil, if its empty; e.g. required to reset defaults for winrm # TODO: move logic into winrm plugin
+    creds[:path] = nil if !creds[:path].nil? && creds[:path].to_s.empty?
 
-    # return the updated config
-    conf
-  end
+    # compat! is available in ruby 2.4+
```
jerryaldrichiii: Could/Should we add a TODO line to change this? If not, any reason for this comment?

lib/train.rb
clintoncwolfe: We're actually modifying the hash in the loop, so we can't use each_key - we must build the keys array, then iterate over that.  Good idea usually, though!

test/unit/helper.rb
```diff
@@ -3,5 +3,6 @@
 require 'minitest/autorun'
 require 'minitest/spec'
 require 'mocha/setup'
+require 'byebug'
```
jerryaldrichiii: Hey! It's in the helper now :smile:

PR #396, opened 2019-01-14
drop ruby 2.0-2.2 support, allow bundler 2.x, bump integration gems, allow mixlib-shellout 3.x

lamont-granquist:
companion to inspec/inspec#3723

this will drop support for chef-client <= 12.13.37 which is  August 15, 2016 and it is not clear that affects anyone practically since inspec only supports >= ruby 2.3 right now.

bonus fix in here for the gcp_test failures -- which are due to the way the ENV var is parsed only once when the require line is first required.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,11 +12,6 @@ if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.2.2')
   gem 'json', '< 2.0'
 end
 
-# bundler is built into Ruby 2.6 and later
-if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.6')
-  gem 'bundler', '>= 1.11'
-end
-
 group :test do
   gem 'minitest', '~> 5.8'
   gem 'rake', '~> 10'
--- a/Gemfile
+++ b/Gemfile
@@ -14,7 +14,7 @@ end
 
 # bundler is built into Ruby 2.6 and later
 if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.6')
-  gem 'bundler', '~> 1.11'
+  gem 'bundler', '>= 1.11'
 end
 
 group :test do
@@ -33,8 +33,8 @@ group :test do
 end
 
 group :integration do
-  gem 'berkshelf', '~> 5.2'
-  gem 'test-kitchen', '~> 1.11'
+  gem 'berkshelf', '>= 6.3.4'
+  gem 'test-kitchen', '>= 1.2.4'
   gem 'kitchen-vagrant'
 end
 
--- a/train.gemspec
+++ b/train.gemspec
@@ -19,11 +19,11 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.0'
+  spec.required_ruby_version = '>= 2.3'
 
   spec.add_dependency 'json', '>= 1.8', '< 3.0'
   # chef-client < 12.4.1 require mixlib-shellout-2.0.1
-  spec.add_dependency 'mixlib-shellout', '~> 2.0'
+  spec.add_dependency 'mixlib-shellout', '>= 2.0'
   spec.add_dependency 'net-ssh', '>= 2.9', '< 6.0'
   spec.add_dependency 'net-scp', '~> 1.2'
   spec.add_dependency 'winrm', '~> 2.0'
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,10 +16,6 @@ branches:
 
 matrix:
   include:
-  - rvm: 2.2.10
-    before_install:
-      - gem install bundler
-      - gem --version
   - rvm: 2.3.8
     before_install:
       - gem install bundler
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,9 +8,6 @@ environment:
   winrm_pass: Pass@word1
 
   matrix:
-    - ruby_version: "22"
-      train_target: winrm://test_user@localhost:5985
-
     - ruby_version: "24"
       train_target: winrm://test_user@localhost:5985
 
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -25,7 +25,7 @@ module Train::Transports
     # https://cloud.google.com/compute/docs/regions-zones/changing-default-zone-region
     # can also specify project via env var:
     option :google_cloud_project, required: false do
-     ENV['GOOGLE_CLOUD_PROJECT']
+      ENV['GOOGLE_CLOUD_PROJECT']
     end
     option :google_super_admin_email, required: false do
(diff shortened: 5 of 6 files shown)


lamont-granquist:
my guess is that the options DSL needs to take syntax something like:

```
option :google_application_credentials, required: false, default: -> { ENV['GOOGLE_APPLICATION_CREDENTIALS'] }
```

and a block/proc/lambda as a default should be lazily evaluated (not sure if the value should get memoized, i had a short argument in my head about it without any clarity -- one way or another probably both behavior should get supported).

lamont-granquist:
rewrote the fix to lazy the default values.  that looks a lot better.

lamont-granquist:
yeah those all should get fixed similarly.  slurping in an ENV var at class loading time is a very brittle pattern.

Gemfile
```diff
@@ -14,7 +14,7 @@ end
 
 # bundler is built into Ruby 2.6 and later
 if Gem::Version.new(RUBY_VERSION) < Gem::Version.new('2.6')
-  gem 'bundler', '~> 1.11'
```
tas50: Do we even need to set a floor here?

train.gemspec
```diff
@@ -19,11 +19,11 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.0'
+  spec.required_ruby_version = '>= 2.3'
 
   spec.add_dependency 'json', '>= 1.8', '< 3.0'
   # chef-client < 12.4.1 require mixlib-shellout-2.0.1
```
tas50: Do we need this comment anymore?

Gemfile
lamont-granquist: i'm not sure it even matters at all?

lamont-granquist: https://github.com/inspec/train/pull/98 is why the floor is there.

lamont-granquist: yeah this clearly doesn't make any sense going forwards since in ruby >= 2.6 this line has to be avoided entirely anyway...

train.gemspec
lamont-granquist: dunno, i don't really understand anything about that comment or what it was important enough of an observation to merit (which meant i didn't want to touch it)

rhass:
:+1:

train.gemspec
```diff
@@ -19,11 +19,11 @@ Gem::Specification.new do |spec|
   spec.test_files    = spec.files.grep(%r{^(test|spec|features)/})
   spec.require_paths = ['lib']
 
-  spec.required_ruby_version = '>= 2.0'
+  spec.required_ruby_version = '>= 2.3'
```
robbkidd: I recommend that this change be omitted from this PR and performed on a separate PR related to removing old Ruby versions which we suspect requires a major version bump.

lamont-granquist:
Keep in mind that current inspec already sets a `required_ruby_version = '>= 2.3'` and that is how people consume train.

We also do not test at all on ruby 2.0 or 2.1 so those are kind of lies.

And for ruby 2.2 the reason why it is being removed is that bundler 2.0 / rubygems 3.0 no longer support it.

lamont-granquist:
gem installing inspec on chef 12.5.1 is already broken:

```
# /opt/chef/embedded/bin/gem install inspec
Fetching: tty-screen-0.6.5.gem (100%)
Successfully installed tty-screen-0.6.5
Fetching: wisper-2.0.0.gem (100%)
Successfully installed wisper-2.0.0
Fetching: tty-reader-0.5.0.gem (100%)
Successfully installed tty-reader-0.5.0
Fetching: timers-4.3.0.gem (100%)
ERROR:  Error installing inspec:
	timers requires Ruby version >= 2.2.1.
```

same results on 12.13.37:

```
# /opt/chef/embedded/bin/gem install inspec
Fetching: train-1.6.3.gem (100%)
Successfully installed train-1.6.3
Fetching: necromancer-0.4.0.gem (100%)
Successfully installed necromancer-0.4.0
Fetching: strings-ansi-0.1.0.gem (100%)
Successfully installed strings-ansi-0.1.0
Fetching: unicode_utils-1.4.0.gem (100%)
Successfully installed unicode_utils-1.4.0
Fetching: strings-0.1.4.gem (100%)
Successfully installed strings-0.1.4
Fetching: tty-table-0.10.0.gem (100%)
Successfully installed tty-table-0.10.0
ERROR:  Error installing inspec:
	timers requires Ruby version >= 2.2.1.
```

It looks like inspec itself dropped support for ruby 2.2 in v1.45.7 from https://github.com/inspec/inspec/commit/f05e22ad3387f7a7fa70a0d56bf6595127bcd739 (released i guess in v1.45.9) and this tty change came in  v3.1.0 from https://github.com/inspec/inspec/commit/c0a46cf8416e7d46abf712d14f08883ac5689842.

lamont-granquist:
so @robbkidd i'm not seeing the point in worrying about this change, the ship pretty much already sailed.

lamont-granquist:
Also not that upgrading rubygems leads you down a path that eventually works:

```
# /opt/chef/embedded/bin/gem update --system 2.7.8
[....snip...]
# /opt/chef/embedded/bin/gem install inspec
ERROR:  Error installing inspec:
	The last version of timers (~> 4.0) to support your Ruby & RubyGems was 4.1.2. Try installing it with `gem install timers -v 4.1.2` and then running the current command again
	timers requires Ruby version >= 2.2.1. The current ruby version is 2.1.0.
# gem install timers -v 4.1.2
Fetching: hitimes-1.3.1.gem (100%)
ERROR:  Error installing timers:
	The last version of hitimes (>= 0) to support your Ruby & RubyGems was 1.3.0. Try installing it with `gem install hitimes -v 1.3.0` and then running the current command again
	hitimes requires Ruby version >= 2.2.2. The current ruby version is 2.1.0.
# gem install hitimes -v 1.3.0
Fetching: hitimes-1.3.0.gem (100%)
Building native extensions. This could take a while...
Successfully installed hitimes-1.3.0
1 gem installed
# gem install timers -v 4.1.2
Fetching: timers-4.1.2.gem (100%)
Successfully installed timers-4.1.2
1 gem installed
# /opt/chef/embedded/bin/gem install inspec
Fetching: tty-prompt-0.18.1.gem (100%)
Successfully installed tty-prompt-0.18.1
Fetching: inspec-3.2.6.gem (100%)
ERROR:  Error installing inspec:
	The last version of inspec (>= 0) to support your Ruby & RubyGems was 1.44.8. Try installing it with `gem install inspec -v 1.44.8`
	inspec requires Ruby version >= 2.3. The current ruby version is 2.1.0.
# gem install inspec -v 1.44.8
Fetching: train-0.32.0.gem (100%)
Successfully installed train-0.32.0
Fetching: rainbow-2.2.2.gem (100%)
Building native extensions. This could take a while...
Successfully installed rainbow-2.2.2
Fetching: inspec-1.44.8.gem (100%)
Successfully installed inspec-1.44.8
3 gems installed
```

clintoncwolfe:
OK, all that said, we're OK to move ahead.

PR #397, opened 2019-01-16
[aws] Use last SDK

marcmillien:
Move to SDK 3 in particular because many other aws libs depends on it, for example the last versions of https://rubygems.org/gems/aws-sdk-ec2 have this dependency: `aws-sdk-core ~> 3, >= 3.39.0`.

This is an alternative solution for https://github.com/inspec/train/pull/398


clintoncwolfe:
Thanks for the contribution, but we have to reject this.

Our plan:
 * Finalise inspec/train-aws as a SDK 3 plugin
 * Remove AWS support form the train gem
 * Add a gemspec dependency on train-aws from inspec.

PR #398, opened 2019-01-16
[aws] SDK 3 and modularity

marcmillien:
Take advantage of aws sdk3 modularity
- reduce the tree of dependencies
- move to last aws sdk version

This is an alternative solution for https://github.com/inspec/train/pull/397

### DIFF
--- a/lib/train/transports/aws.rb
+++ b/lib/train/transports/aws.rb
@@ -1,6 +1,6 @@
 # encoding: utf-8
 require 'train/plugins'
-require 'aws-sdk'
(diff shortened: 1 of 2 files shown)


clintoncwolfe:
Thanks for the contribution, but we have to reject this.

Our plan:
 * Finalise inspec/train-aws as a SDK 3 plugin
 * Remove AWS support form the train gem
 * Add a gemspec dependency on train-aws from inspec.

marcmillien:
Ok thanks :+1:

PR #401, opened 2019-01-25
Adds support for IOS XE virtual appliances too

TheLonelyGhost:
Tested manually against an appliance I'm about to not have access to anymore. Not sure how to add test support for this change quite yet, but I've saved off a copy of `show version` to a local text file in case it involves stubbing a command. Assistance?

Target platform is a Cisco IOS XE Virtual appliance v16.9.1 (Fuji).

### DIFF
--- a/lib/train/platforms/detect/helpers/os_common.rb
+++ b/lib/train/platforms/detect/helpers/os_common.rb
@@ -79,6 +79,11 @@ module Train::Platforms::Detect::Helpers
         return @cache[:cisco] = { version: m[2], model: m[1], type: 'ios-xe' }
       end
 
+      m = res.match(/Cisco IOS Software \[([^,]+)\], Virtual XE Software [^,]+, Version (\d+\.\d+\.\d+[A-Z]*)/)
+      unless m.nil?
+        return @cache[:cisco] = { version: m[2], model: m[1], type: 'ios-xe' }
+      end
+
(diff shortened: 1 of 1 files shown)


lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -79,6 +79,11 @@ def cisco_show_version
         return @cache[:cisco] = { version: m[2], model: m[1], type: 'ios-xe' }
       end
 
+      m = res.match(/Cisco IOS Software \[([^,]+)\], Virtual XE Software [^,]+, Version (\d+\.\d+\.\d+[A-Z]*)/)
```
jerryaldrichiii: I can't decide if this is different enough that I wouldn't just modify the other regex for Cisco IOS XE. :thinking:

I think this is fine though. Maybe add a comment to show this is Virtual XE?

TheLonelyGhost: Yeah, I also couldn't decide. I opted for a separate case for it since it's much easier to combine regexes than to take them apart. I'll add in comments to indicate why it's separate. 👍

clintoncwolfe:
As this has been untouched for a while, and requires tests and a special testing environment, I'm going to close this for now.  If you have time to re-visit it and have access to the test environment, please do re-open.  Thanks!

PR #423, opened 2019-03-18
Add data callbacks to WinRM and SSH

marcparadise:
In order to support existing bootstrap functionality for `knife
bootstrap`s conversion to Train, clients need a method to receive
data from the target host as it is being generated.

This callback will be invoked for SSH when data arrives on the channel,
and for WinRM when data is returned on stdout.


marcparadise:
Switching base branch on github has broken this - appveyor build won't run. I'll re-submit clean.

PR #425, opened 2019-03-18
[SUSTAIN-955] Add data callbacks to WinRM and SSH

marcparadise:
In order to support existing bootstrap functionality for `knife
bootstrap`s conversion to Train, clients need a method to receive
data from the target host as it is being generated.

This callback will be invoked for SSH when data arrives on the channel,
and for WinRM when data is returned on stdout.

### DIFF
--- a/lib/train/plugins/base_connection.rb
+++ b/lib/train/plugins/base_connection.rb
@@ -114,10 +114,14 @@ class Train::Plugins::Transport
 
     # This is the main command call for all connections. This will call the private
     # run_command_via_connection on the connection with optional caching
-    def run_command(cmd)
-      return run_command_via_connection(cmd) unless cache_enabled?(:command)
+    #
+    # This command accepts an optional data handler block. When provided,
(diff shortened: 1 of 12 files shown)


marcparadise:
@jerryaldrichi I have added some tests.  Though in the he case of SSH, I wonder if there's value in it - E_TOO_MUCH_MOCKING.

marcparadise:
@clintoncwolfe I've updated this to use blocks from your suggestion in the review meeting. That's expanded the scope of the change a bit, but it is a bit cleaner - and has the advantage of not invoking the callback for internal train commands used for platform detection.   Callback handling has been added to winrm and ssh; if we need it for other transports, we can have separate implementation PRs for those.

marcparadise:
The failing test is a transient failure that I saw locally, it seems to depend on the test case randomization sequence.

clintoncwolfe:
@marcparadise could you take care of the merge conflict? After that I think we're ready to merge.

clintoncwolfe:
Adopted on #431 to fix merge conflict

PR #430, opened 2019-03-20
Silence verify_host_key warning from net-ssh

clintoncwolfe:
The SSH transport always wants to disable host key verification. In net-ssh 5+, the acceptable values for verify_host_key changed ;we currently set it to `false`, which now results in a deprecation warning from `net-ssh`
```
verify_host_key: false is deprecated, use :never
```

We already choose what option name to use based on the net-ssh version; this PR now chooses the value dynamically, as well.

As I found when passing values from a config file, there is a good chance the values will come in as Strings, so the PR does some questionable remapping of things to try to catch the common cases.

Signed-off-by: Clinton Wolfe <clintoncwolfe@gmail.com>

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -199,7 +199,6 @@ module Train::Transports
     def verify_host_key_value(given)
       current_net_ssh = Net::SSH::Version::CURRENT
       new_value_version = Net::SSH::Version[5, 0, 0]
-
       if current_net_ssh >= new_value_version
         # 5.0+ style
         {
@@ -217,6 +216,7 @@ module Train::Transports
         {
           'true' => true,
           'false' => false,
+          nil => false,
         }.fetch(given, given)
       end
     end
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -210,6 +210,7 @@ module Train::Transports
           # May be correct value, but strings from JSON config
           'always' => :always,
           'never' => :never,
+          nil => :never,
         }.fetch(given, given)
       else
         # up to 4.2 style
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -167,9 +167,9 @@ module Train::Transports
         non_interactive:        opts[:non_interactive],
         transport_options:      opts,
       }
-      # disable host key verification. The hash key to use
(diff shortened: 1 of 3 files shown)


lib/train/transports/ssh.rb
```diff
@@ -167,9 +167,9 @@ def connection_options(opts)
         non_interactive:        opts[:non_interactive],
         transport_options:      opts,
       }
-      # disable host key verification. The hash key to use
+      # disable host key verification. The hash key and value to use
       # depends on the version of net-ssh in use.
-      connection_options[verify_host_key_option] = opts[:verify_host_key] || false
+      connection_options[verify_host_key_option] = opts[:verify_host_key]
```
clintoncwolfe: Note that the `||` default logic is now in the `option` call on line 66.

lib/train/transports/ssh.rb
```diff
@@ -193,6 +193,34 @@ def verify_host_key_option
...
+      new_value_version = Net::SSH::Version[5, 0, 0]
+      if current_net_ssh >= new_value_version
+        # 5.0+ style
+        {
+          # It's not a boolean anymore.
+          'true' => :always,
+          'false' => :never,
+          true => :always,
+          false => :never,
+          # May be correct value, but strings from JSON config
```
marcparadise: Missing the nil case here, since it's possible for someone to specify the option w/ nil value.

lib/train/transports/ssh.rb
```diff
@@ -193,6 +193,34 @@ def verify_host_key_option
       current_net_ssh >= new_option_version ? :verify_host_key : :paranoid
     end
 
+    # Likewise, version <5 accepted false; 5+ requires :never or will
+    # issue a deprecation warning. This method allows a lot of common
+    # things through.
+    def verify_host_key_value(given)
```
marcparadise: Should this have tests?

lib/train/transports/ssh.rb
clintoncwolfe: Done, thanks!

lib/train/transports/ssh.rb
clintoncwolfe: I've added some, but stubbing the const used to detect the net-ssh version (which affects the mapping, here) proved elusive. So, the tests check whatever net-ssh version is currently being used, which is not ideal.

PR #435, opened 2019-04-17
 Require ed25519 and bcrypt_pbkdf gems for ed25519 support 

tas50:
Modern distros are switching to ed25519 ssh keys and we need to make
sure train supports this out of the box.

I also shuffled the deps around so we can easily diff the train and train-core gemspecs.

### DIFF
--- a/train-core.gemspec
+++ b/train-core.gemspec
@@ -30,10 +30,12 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
+  spec.add_dependency 'json', '>= 1.8', '< 3.0'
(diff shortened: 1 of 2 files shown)


train-core.gemspec
```diff
@@ -30,10 +30,12 @@ Gem::Specification.new do |spec|
   spec.executables   = spec.files.grep(%r{^bin/}) { |f| File.basename(f) }
   spec.require_paths = ['lib']
 
+  spec.add_dependency 'json', '>= 1.8', '< 3.0'
+  spec.add_dependency 'mixlib-shellout', '>= 2.0', '< 4.0'
   spec.add_dependency 'net-ssh', '>= 2.9', '< 6.0'
```
tas50: This is left at 3.x since the ed25519 gem does not directly depend on net-ssh. If a person uses the old net-ssh they won't get ed25519 support, but nothing will break.

PR #436, opened 2019-04-22
Add missed cisco_ios_connection to train-core.

marcparadise:
This is required in transports/ssh.rb, so
needs to be included in the gem.


clintoncwolfe:
I really wish we didn't have to do this; it digs a deeper hole.  I have opened #437 to try to decouple SSH and Cisco one day.  For today, we need to let this ride.

PR #443, opened 2019-04-26
The Windows release also changed..

miah:
Signed-off-by: Miah Johnson <miah@chia-pet.org>

### DIFF
--- a/test/windows/winrm_test.rb
+++ b/test/windows/winrm_test.rb
@@ -27,9 +27,9 @@ describe 'windows winrm command' do
 
   it 'verify os' do
     os = conn.os
-    os[:name].must_equal 'windows_server_2016_datacenter'
+    os[:name].must_equal /windows_server.*_datacenter/
     os[:family].must_equal 'windows'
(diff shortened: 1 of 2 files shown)


test/windows/winrm_test.rb
```diff
@@ -27,9 +27,9 @@
 
   it 'verify os' do
     os = conn.os
-    os[:name].must_equal 'windows_server_2012_r2_datacenter'
+    os[:name].must_equal 'windows_server_2016_datacenter'
     os[:family].must_equal 'windows'
-    os[:release].must_equal '6.3.9600'
+    os[:release].must_equal '10.0.14393'
```
zenspider: I'm fine with these... but it might be more robust for us if we use regexps instead... `/windows_server.*_datacenter/` and `/\d+(\.\d+)+/` or something similar. We don't care about the specific value, we care that the value is there and sane looking.

miah: Thanks, I updated the PR with those regular expressions. I agree this will be a problem any time the image changes when all we really want is to ensure Train properly detects the image details.

PR #448, opened 2019-05-10
Remove WinRM support in favor of train-winrm plugin

clintoncwolfe:
This PR removes WinRM support from the core of train, in favor of using it via the train-winrm plugin. Be sure to update your dependencies to include train-winrm.

~~As one of the major themes of Train 2 is pushing transports into plugins for lighter packaging, we're only doing a minor version bump on this.~~

As this will break any user of train that relies on winrm, good semver discipline would be to do a major version bump.


lib/train/platforms/detect/specifications/os.rb
```diff
@@ -21,7 +21,7 @@ def self.load
             is_windows = false
             is_windows = true if winrm?
 
-            if @backend.class.to_s == 'Train::Transports::Local::Connection'
+            if backend_name == 'Train::Transports::Local::Connection'
```
zenspider: Did you figure out where this methed is coming in from? I only see the above being included via Scanner and I get lost after that.

test/windows/local_test.rb
```diff
@@ -7,22 +7,23 @@
...
+    target_config = Train.target_config({logger: Logger.new(STDERR, level: :info)})
     # initialize train
     backend = Train.create('local', target_config)
   end
   let(:conn) { backend.connection }
 
   it 'verify os' do
     os = conn.os
-    os[:name].must_equal 'windows_server_2012_r2_datacenter'
+    os[:name].must_match(/windows_server.*/)
```
zenspider: 👍

test/unit/platforms/detect/os_common_test.rb
```diff
@@ -16,26 +16,14 @@ def initialize
...
-      detector.instance_variable_set(:@backend, be)
-      detector.winrm?.must_equal(false)
-    end
-
     it 'return winrm? false when winrm is not loaded' do
-      require 'train/transports/winrm'
-      winrm = Train::Transports.send(:remove_const, 'WinRM')
       be = mock('Backend')
       detector.instance_variable_set(:@backend, be)
       detector.winrm?.must_equal(false)
```
zenspider: Why not set up this test similar to the `true` case in order to make it obvious that it is just the backend_name that's really causing `winrm?` to return true vs false? Symmetry is king in these sorts of tests.

clintoncwolfe: Fine

train-core.gemspec
```diff
@@ -37,6 +37,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'mixlib-shellout', '>= 2.0', '< 4.0'
   spec.add_dependency 'net-ssh', '>= 2.9', '< 6.0'
   spec.add_dependency 'net-scp', '>= 1.2', '< 3.0'
-  spec.add_dependency 'winrm', '~> 2.0'
-  spec.add_dependency 'winrm-fs', '~> 1.0'
+  spec.add_dependency 'ed25519', '~> 1.2' # ed25519 ssh key support
```
tas50: If we add these then we need compilation tools in train.

clintoncwolfe: Yes, I agree; but those came in from a rebase.  I'll yank them here.

miah:
This one needs a rebase.

PR #449, opened 2019-05-14
Fixes Uninitialized constant error Train::Transports::SSHFailed

Vasu1105:
Signed-off-by: Vasu1105 <vasundhara.jagdale@msystechnologies.com>

This will fix https://github.com/chef/chef/issues/8533 partially.

### DIFF
--- a/lib/train.rb
+++ b/lib/train.rb
@@ -7,6 +7,7 @@ require 'train/options'
 require 'train/plugins'
 require 'train/errors'
 require 'train/platforms'
+require 'train/transports'
 require 'uri'
 
 module Train
--- a/lib/train/transports.rb
+++ b/lib/train/transports.rb
@@ -1,0 +1,4 @@
+require 'train/transports/ssh'
+
+module Train::Transports
+end


marcparadise:
The transport files are kept pretty isolated from the main application by design.  It would help to have a little more background on this. I looked at the issue you referenced, but didn't see the connection since that was an SSL error.

Vasu1105:
> The transport files are kept pretty isolated from the `train.rb` by design. It would help to have a little more background on this. I looked at the issue you referenced, but didn't see the connection since that was an SSL error.

So If you look at my main issue which is opened in Chef it throws OpenSSL Error and it calls out the method `connect` here  https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L599 which calls out the https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L598 where it throws `Openssl` error and When since raise line here https://github.com/chef/chef/blob/master/lib/chef/knife/bootstrap.rb#L59e not able to load class it give `Unintialized Constan`t error

marcparadise:
Let's fix this on the chef side by requiring the file before we reference the exception, or filtering out exceptions based on what the loaded protocol is. Train is doing the right thing here - the SSH protocol is never loaded, so it makes sense that when we try to rescue a Train::Transports::SSHFailed error, we should be loading that ourselves.,

PR #450, opened 2019-05-14
Rename ca_trust_file to ca_trust_path

marcparadise:
WinRM expects `ca_trust_path` not `ca_trust_file`


marcparadise:
I've updated this with a check of default, but I couldn't find a meaningful way to verify that it's getting passed through to WinRMConnection - the only locations to validate it are private.

marcparadise:
In the category of 'iffy is better than nothing'... I added validations for the configuration we're passing into `WinRM::Connection`, using `winrm_transport_inst.send` to get the value since `connection_options` is private.

Issue #377, opened 2018-11-03
OpenSUSE Leap platform detection


clintoncwolfe:
### Description

A user reported an issue in which OpenSUSE Leap is not being detected.  

### Train and Platform Version

InSpec 3.0.12
Train 1.5.??
Target: OpenSUSE Leap 15 (Docker)

### Replication Case

Extensive notes and repro instructions on the original issue, https://github.com/inspec/inspec/issues/3574

### Possible Solutions

Implement or refine SUSE detection

megamorf:
I can see that Train's entire SUSE related detection is based around the existence of `/etc/SuSE-release`: 
https://github.com/inspec/train/blob/03afd4dbb990918bb5443b90b1cb49d3fdaf584f/lib/train/platforms/detect/specifications/os.rb#L225-L242

`/etc/os-release` should be checked as a fallback (see the related issue for its contents) since the official openSUSE Leap Docker image has no `/etc/SuSE-release` file.

tas50:
This also means SLES 15 won’t detect correctly. I’m fixing this on the Ohai side and I can fix it here if no one minds.

megamorf:
@tas50 I've created a corresponding ohai issue yesterday: https://github.com/chef/ohai/issues/1289
My ruby skills are too bad to fix it myself in a reasonable time so your assistance would be appreciated :-)

n-rodriguez:
I think the regex is wrong.

The file `/etc/SuSE-release` exists on openSUSE Leap 15 and its content is :
 
```sh
kitchen@b59935020e63:~> cat /etc/SuSE-release 
NAME="openSUSE Leap"
VERSION="15.0"
ID="opensuse-leap"
ID_LIKE="suse opensuse"
VERSION_ID="15.0"
PRETTY_NAME="openSUSE Leap 15.0"
ANSI_COLOR="0;32"
CPE_NAME="cpe:/o:opensuse:leap:15.0"
BUG_REPORT_URL="https://bugs.opensuse.org"
HOME_URL="https://www.opensuse.org/"
```
Here the regex : 

```ruby

tas50:
I'd suggest in the long run that train moves away from using files like /etc/SuSe-release and instead use /etc/os-release, which exists on every systemd based Linux distro and many distros prior to systemd. It has a standard format and can easily be combined with some case statements to make platform detection much more reliable. I made the switch in Ohai for Chef 15: https://github.com/chef/ohai/pull/1299

[...]

PR #451, opened 2019-05-15
Fix fallback regex for OpenSuse

n-rodriguez:
See https://github.com/inspec/train/issues/377

This regex matches both case :

`VERSION = 15.0` and 
`VERSION="15.0"`

Thank you!

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -237,8 +237,10 @@ module Train::Platforms::Detect::Specifications
       plat.family('suse').in_family('linux')
           .detect {
             if !(suse = unix_file_contents('/etc/SuSE-release')).nil?
+              # https://rubular.com/r/UKaYWolCYFMfp1
               version = suse.scan(/VERSION = (\d+)\nPATCHLEVEL = (\d+)/).flatten.join('.')
-              version = suse[/VERSION = ([\d\.]{2,})/, 1] if version == ''
+              # https://rubular.com/r/b5PN3hZDxa5amV
+              version = suse[/VERSION\s?=\s?"?([\d\.]{2,})"?/, 1] if version == ''
               @platform[:release] = version
               true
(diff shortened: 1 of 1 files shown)


tas50:
@n-rodriguez Can you add a comment above the regex with a permalink from rubular. https://rubular.com/

That way we can have a link in the code to show the various things we're trying to match instead of having to track things back to this PR.

Also you'll need to sign this commit off for the DCO before we can merge it. https://github.com/chef/chef/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

n-rodriguez:
@tas50 done! Thank you!

robbkidd:
Seems OK with more variants: https://rubular.com/r/b5PN3hZDxa5amV

And an example for the regex appearing above it: https://rubular.com/r/UKaYWolCYFMfp1

n-rodriguez:
@robbkidd done! thank you! :+1:

zenspider:
I dunno... should I be the anal-retentive regexp-er? I am not sure it actually matters because this is coming from some vendored source?

I'm gonna approve, but just so you know, that `\s` accepts things like tabs or newlines. You probably just want `<space>?`... and the `"?` allows for the quotes to be on one side and not the other. Again, probably doesn't matter, but in other contexts that sort of thing is crucial.

n-rodriguez:
I'm gonna change \s to space

zenspider:
I meant `/ ?/` (added slashes so you can see the space, vs `/\s?/`. The first one only accepts a space.

n-rodriguez:
> I meant `/ ?/` (added slashes so you can see the space, vs `/\s?/`. The first one only accepts a space.

Ok. Do you want me to change it?

zenspider:
I'll get it. Just consider it for next time.

n-rodriguez:
> I'll get it. Just consider it for next time.

I'll remember! Thanks!

PR #452, opened 2019-05-15
Fix failing unit tests verify_host_key in ssh

marcparadise:
These tests would pass/fail based on your installed version of Net::SSH
because the expected behavior is different for >= 5.

I updated the tests to account for the different behaviors.

### DIFF
--- a/test/unit/transports/ssh_test.rb
+++ b/test/unit/transports/ssh_test.rb
@@ -51,18 +51,6 @@ describe 'ssh transport' do
       connection_options.key?(:paranoid).must_equal false
     end
 
-    it 'defaults verify_host_key option to false, and does not set paranoid' do
-      connection_options[:verify_host_key].must_equal false
-      connection_options.key?(:paranoid).must_equal false
-    end
-
-    describe "when caller sets verify_host_key in options" do
-      let(:opts) { { verify_host_key: true } }
-      it 'the provided value is used instead of the default' do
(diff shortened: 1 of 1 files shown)


test/unit/transports/ssh_test.rb
```diff
@@ -46,23 +46,38 @@
...
+    end
+
+    describe 'when caller sets verify_host_key to an unknown value in options' do 
+      let(:opts) { { verify_host_key: "moonshine"} }
+      it 'it keeps the provided value' do
+        connection_options[:verify_host_key].must_equal "moonshine"
       end
     end
 
+
```
tas50: This is probably going to fail rubocop

test/unit/transports/ssh_test.rb
```diff
@@ -46,23 +46,38 @@
...
+    }
 
     it 'does not set a paranoid option - deprecated in net-ssh 4.2' do
       connection_options.key?(:paranoid).must_equal false
     end
 
     it 'defaults verify_host_key option to false, and does not set paranoid' do
-      connection_options[:verify_host_key].must_equal false
+      expected = old_net_ssh ? false : :never
+      connection_options[:verify_host_key].must_equal expected
```
marcparadise: Now that I'm re-reading this, I'm thinking separate tests that force the version behavior and verify each separately would be better.

PR #458, opened 2019-05-23
Minor cleanup to regexp for correctness.

zenspider:
I'm just being a tad anal retentive here. I don't want the quotes to
be mis-matched and I don't want the \s to match NLs.

Signed-off-by: Ryan Davis <zenspider@chef.io>

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -239,8 +239,8 @@ module Train::Platforms::Detect::Specifications
           unless (suse = unix_file_contents("/etc/SuSE-release")).nil?
             # https://rubular.com/r/UKaYWolCYFMfp1
             version = suse.scan(/VERSION = (\d+)\nPATCHLEVEL = (\d+)/).flatten.join(".")
-            # https://rubular.com/r/b5PN3hZDxa5amV
-            version = suse[/VERSION\s?=\s?"?([\d\.]{2,})"?/, 1] if version == ""
+            # https://rubular.com/r/xegcPXKEXJrJl5
+            version = suse[/VERSION *= *("?)([\d\.]{2,})\1$/, 2] if version == ""
             @platform[:release] = version
             true
           end


lib/train/platforms/detect/specifications/os.rb
```diff
@@ -237,10 +237,8 @@ def self.load
       plat.family('suse').in_family('linux')
           .detect {
             if !(suse = unix_file_contents('/etc/SuSE-release')).nil?
-              # https://rubular.com/r/UKaYWolCYFMfp1
               version = suse.scan(/VERSION = (\d+)\nPATCHLEVEL = (\d+)/).flatten.join('.')
-              # https://rubular.com/r/b5PN3hZDxa5amV
-              version = suse[/VERSION\s?=\s?"?([\d\.]{2,})"?/, 1] if version == ''
+              version = suse[/VERSION *= *("?)([\d\.]{2,})\1$/, 2] if version == ''
```
clintoncwolfe: I find this harder to read (` *= *` almost reads as an operator), and you've removed a helpful comment.  Please have some mercy on your fellow developers, and keep/update explanatory comments.

miah: I don't think your intention was to imply malice, but 'please have some mercy' reads kinda harshly.

zenspider: First, I agree with @miah here. There’s another PR comment with “be kind and ...” that also reads like you’re implying malicious intent. I don’t think you are, but still... this is a public forum.

Second, The rubular comments seem very inconsistently applied to our projects... tho I can now see that they’re used more in ohai & chef, but it’s still only 18 references across 12 files (out of 3358 files).

Third, I personally don’t think this regexp is that hard to follow as far as our regexps go. I’ll add it back if you want, but I’d rather do a brown-bag (or whatever the remote equivalent would be) on regexps instead. Miah has already said that would help her and I would be happy to do it.

Finally… For fun, I wrote a quick script to find and “score” all our regexps and … well… let’s just say that there are going to be some more issues filed shortly as it is exposing some problems.

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -239,8 +239,8 @@ def self.load
           unless (suse = unix_file_contents("/etc/SuSE-release")).nil?
             # https://rubular.com/r/UKaYWolCYFMfp1
             version = suse.scan(/VERSION = (\d+)\nPATCHLEVEL = (\d+)/).flatten.join(".")
-            # https://rubular.com/r/b5PN3hZDxa5amV
-            version = suse[/VERSION\s?=\s?"?([\d\.]{2,})"?/, 1] if version == ""
+            # https://rubular.com/r/xegcPXKEXJrJl5
```
tas50: 😄

PR #460, opened 2019-05-24
Add codeclimate setup.

miah:
This updates the codeclimate setup that we use in InSpec/Train.


zenspider:
Rebase this. If it is green, we should merge.

PR #470, opened 2019-07-03
Resolve chefstyle warnings in 0.13

tas50:
In 0.13 we upgraded to Rubocop engine from 0.62 to 0.72. This resolved a
ton of bugs like this.


zenspider:
Relatedly... what do we need to do to get the build kite stuff working?

PR #476, opened 2019-07-11
Use /etc/os-release for SUSE detection

russellcardullo:
## Description

The file `/etc/SuSE-release` was deprecated in older SUSE versions and as of SUSE/SLES 15.X is no longer present.  This switches the detection to use the values in `/etc/os-release`.

## Related Issue

#377 

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -236,22 +236,18 @@ module Train::Platforms::Detect::Specifications
       # suse family
       plat.family("suse").in_family("linux")
           .detect do
-            if !(suse = unix_file_contents("/etc/SuSE-release")).nil?
-              # https://rubular.com/r/UKaYWolCYFMfp1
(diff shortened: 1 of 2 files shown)


tas50:
/etc/os-release is definitely the preferred detection place, but I believe SLES 11 doesn't have this file so this would break SLES 11 detection.

russellcardullo:
@tas50 thanks, agreed we definitely shouldn’t break versions without that file.  I think I can add back the old detection as a fallback and will test that out.

tas50:
Take a look at how we're doing this in Ohai. It starts with /etc/os-release and falls back to all the old configs. https://github.com/chef/ohai/blob/master/lib/ohai/plugins/linux/platform.rb

russellcardullo:
@tas50 I added the fallback code back in which will get used when `/etc/os-release` is not present.  I looked as far back as SUSE 9.X and the format of those files should work with the code here, but let me know if there are any other issues that could show up.

clintoncwolfe:
Thanks @russellcardullo for this PR! I've adopted it as #505, and performed a rebase - should be merged shortly. Thanks again!

PR #484, opened 2019-07-23
Add Windows to the verify pipeline

miah:


Gemfile
```diff
@@ -11,6 +11,8 @@ group :test do
   gem "concurrent-ruby", "~> 1.0"
   gem "pry-byebug"
   gem "m"
+  gem "ed25519" # ed25519 ssh key support
+  gem "bcrypt_pbkdf" # ed25519 ssh key support
```
miah: I moved this because I want to make sure `ed25519` and `bcrypt_pbkdf` are still present during testing but don't want the Integration group during verification testing.

Issue #506, opened 2019-08-12
Wrong raise error method arguments


vsingh-msys:
# Version:
3.0.1

# Environment:
Host: Ubuntu 18.04
Target node: Centos 6.5
Protocol: SSH

# Scenario:
By default CentOS 6 is configured with `requiretty`
While hitting ssh with `sudo: true` getting backtrace error instead of a proper error message.

# Steps to Reproduce:
- SSH ready  any machine with(configure `requiretty`) or CentOS 6.5(default configured)
- Run 
```
train = Train.create('ssh', host: 'IP_ADDRESS', port: 22, user: 'user', password: 'password', sudo: true)
train.connection
```
# Expected Result:
```
Traceback (most recent call last):
	5: from ./test.rb:9:in `<main>'
	4: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:82:in `connection'
	3: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:238:in `create_new_connection'
	2: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:238:in `new'
	1: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh_connection.rb:53:in `initialize'
/.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/extras/command_wrapper.rb:173:in `load': Sudo failed: Sudo requires a TTY. Please see the README on how to configure sudo to allow for non-interactive usage. (Train::UserError)
```
# Actual Result:
```
Traceback (most recent call last):
	7: from ./test.rb:9:in `<main>'
	6: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:82:in `connection'
	5: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:238:in `create_new_connection'
	4: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh.rb:238:in `new'
	3: from /.chefdk/gem/ruby/2.6.0/gems/train-3.0.1/lib/train/transports/ssh_connection.rb:53:in `initialize'
[...]

PR #508, opened 2019-08-12
Fix raise invalid arguments

vsingh-msys:
Fixed improper error msg raised  `backtrace must be Array of String (TypeError)`

Before Fix: 
```
backtrace must be Array of String (TypeError)
```

After Fix: 

```
Sudo failed: Sudo requires a TTY. Please see the README on how to configure sudo to allow for non-interactive usage. (Train::UserError)
```

## Description 
 - Kernel#raise takes caller string array as a third argument
 - But it appeared `reason` String passed while raising an error.
 - Reference https://ruby-doc.org/core-2.6.3/Kernel.html#method-i-raise

## Related Issue
https://github.com/inspec/train/issues/506

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -169,8 +169,8 @@ module Train::Extras
         res = LinuxCommand.new(transport, options)
         verification_res = res.verify
         if verification_res
-          msg, reason = verification_res
-          raise Train::UserError, "Sudo failed: #{msg}", reason
+          msg, _reason = verification_res
+          raise Train::UserError, "Sudo failed: #{msg}"
         end
         res
       elsif transport.platform.windows?


lib/train/extras/command_wrapper.rb
```diff
@@ -169,8 +169,8 @@ def self.load(transport, options)
         res = LinuxCommand.new(transport, options)
         verification_res = res.verify
         if verification_res
-          msg, reason = verification_res
-          raise Train::UserError, "Sudo failed: #{msg}", reason
+          msg, _reason = verification_res
+          raise Train::UserError, "Sudo failed: #{msg}"
```
zenspider: Wow. Good find... but that `verify` method is just a mess and should be cleaned up (in a different PR). Returning `nil` or a string or an array of a string or a symbol is _not right_. I'll file a separate issue against that. Thank you.

vsingh-msys: @zenspider, will cleanup verify method thanks!

PR #512, opened 2019-08-29
Let expeditor respond to pull request

miah:
closes https://github.com/inspec/inspec/issues/4429


tas50:
Totally snagging this for all the other projects

PR #515, opened 2019-09-06
Fix inspec detect on SLES

christian-wtd:
This is a fix for the broken remote execution of Inspec on SUSE SLES 12.

## Description
More details can be found in https://github.com/inspec/inspec/issues/4384

Basically the problem appears for non-root users with sudosh blocking OS detection without sudo and blocks any inspec run for remote operations. Neither the /etc/os-release file is readable nor the uname command is executable without sudo in such an environment.

This has been verified working with Inspec CLI and also Inspec Library Runner Class.

## Related Issue
https://github.com/inspec/inspec/issues/4384

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/platforms/detect/helpers/os_common.rb
+++ b/lib/train/platforms/detect/helpers/os_common.rb
@@ -24,7 +24,6 @@ module Train::Platforms::Detect::Helpers
       return @files[path] if @files.key?(path)
 
       res = @backend.run_command("test -f #{path} && cat #{path}")
-
       # ignore files that can't be read
       @files[path] = res.exit_status == 0 ? res.stdout : nil
       @files[path]
--- a/lib/train/platforms/detect/helpers/os_common.rb
+++ b/lib/train/platforms/detect/helpers/os_common.rb
@@ -24,7 +24,7 @@ module Train::Platforms::Detect::Helpers
       return @files[path] if @files.key?(path)
 
       res = @backend.run_command("test -f #{path} && cat #{path}")
-      
+
       # ignore files that can't be read
       @files[path] = res.exit_status == 0 ? res.stdout : nil
       @files[path]
@@ -49,6 +49,7 @@ module Train::Platforms::Detect::Helpers
 
     def unix_uname_s
       return @uname[:s] if @uname.key?(:s)
+
       @uname[:s] = command_output("uname -s")
     end
 
@@ -60,7 +61,7 @@ module Train::Platforms::Detect::Helpers
 
     def unix_uname_m
       return @uname[:m] if @uname.key?(:m)
-      
+
       @uname[:m] = command_output("uname -m")
     end
 
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -287,7 +287,7 @@ class Train::Transports::SSH
         # wrap commands if that is configured
         if !@cmd_wrapper.nil?
           cmd = @cmd_wrapper.run(cmd)
-        # else check if sudo is configured so we assume Linux is at least the platform so we use Linux wrapper 
+        # else check if sudo is configured so we assume Linux is at least the platform so we use Linux wrapper
         elsif @transport_options[:sudo]
           cmd = LinuxCommand.new(self, @transport_options).run(cmd)
         end
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -287,8 +287,8 @@ class Train::Transports::SSH
         # wrap commands if that is configured
         if !@cmd_wrapper.nil?
           cmd = @cmd_wrapper.run(cmd)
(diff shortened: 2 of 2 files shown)


christian-wtd:
Hello,

did anyone already had a chance to look at this?

Best regards
Christian

lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -23,7 +23,11 @@ def unix_file_contents(path)
       # keep a log of files incase multiple checks call the same one
       return @files[path] if @files.key?(path)
 
-      res = @backend.run_command("test -f #{path} && cat #{path}")
+      if ARGV.include? "--sudo"
```
clintoncwolfe: Hrm, I'd like to find a better way of doing that - reaching back to `ARGV` here feels bad and brittle. Do we have access to the Connection options anywhere, perhaps within the `@backend` object?

lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -49,19 +53,31 @@ def command_output(cmd)
     def unix_uname_s
       return @uname[:s] if @uname.key?(:s)
 
-      @uname[:s] = command_output("uname -s")
+      if ARGV.include? "--sudo"
```
clintoncwolfe: This is pretty repetitive - perhaps we could make a method, `command_prefix`, which conditionally returns either `"sudo"` or `""`, and then all these conditionals would just look like 
```
@uname[:s] = command_output("#{command_prefix} uname -s")
```

christian-wtd: Ok, actually I do see where we have the problem visible in the stack trace:
        27: from lib/train/transports/ssh_connection.rb:53:in `initialize'
        26: from /lib/train/extras/command_wrapper.rb:166:in `load'
at this point we initialize the CommandWrapper in
/lib/train/transports/ssh_connection.rb:
`53      @cmd_wrapper            = CommandWrapper.load(self, @transport_options)`

meanwhile deeper in the stack we find:
/lib/train/transports/ssh_connection.rb:284:in `block in execute_on_channel'

so... this call is run in the scope where we yet not defined @cmd_wrapper because we just try to do, resulting in:
`283        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?`

to not add sudo because we don't have the wrapper yet because we are just loading it.

So in fact there may be a workaround required most probably at this point.

Let me think about that some more time and I may find a proper solution :-)

christian-wtd:
Ok. now I have fucked up GIT somehow :-(
there should be only one cimmt left for the 
lib/train/transports/ssh_connection.rb
but I mixed it up.

christian-wtd:
So having the different approach now in the correct file we should be good to go and also we no longer use ARGV.
Single place, less code and more appropriate place.

christian-wtd:
the reference I made was wrong, should have pointed to my other pull request.

lib/train/transports/ssh_connection.rb
```diff
@@ -285,7 +285,12 @@ def execute_on_channel(cmd, &data_handler)
       exit_status = nil
       session.open_channel do |channel|
         # wrap commands if that is configured
-        cmd = @cmd_wrapper.run(cmd) unless @cmd_wrapper.nil?
+        if !@cmd_wrapper.nil?
+          cmd = @cmd_wrapper.run(cmd)
+        elsif @transport_options[:sudo] || @transport_options[:shell]
+          # if sudo or shell is in the options we can assume Linux/Unix
+          cmd = LinuxCommand.new(self, @transport_options).run(cmd)
```
clintoncwolfe: For other reviewers, your choices are LinuxCommand (which implements sudo facilities), WindowsCommand, or a shared superclass. LinuxCommand is too tightly named, but this PR appears to be using it correctly.

christian-wtd:
Any ETA when this gets merged and released?

lib/train/transports/ssh_connection.rb
christian-wtd: Effectively this is what other code parts supply, don't want to change lots of other code but yeah, "LinuxCommand" covers all kind of Linux/Unix OS so it should be probably named "UnixCommand".

christian-wtd:
@miah & @clintoncwolfe :
It's nearly 2 months after approval, when this will be merged and released?

christian-wtd:
Thank you very much!

zenspider:
This PR might need to be rolled back as it is causing severe problems in other areas (eg #538) and it is too hard to debug. It should probably also have had a test added before it was merged.

Further, I'm running sles 12 in a docker container and both `uname` and `/etc/os-release` are fine without sudo, so I'm really not sure what this is solving yet.

Finally, given the 9 commits vs the final diff, this should have been squashed or interactive rebased and cleaned up.

zenspider:
This has been addressed via #544.

PR #531, opened 2019-11-06
Update google-api-client version.

skpaterson:
Note that knife-google will also be affected by this change - separate PR will be opened there.  We should not merge this PR until confirmation the version upgrade is ok. 


skpaterson:
Related PRs listed below:
- kitchen-google - https://github.com/test-kitchen/kitchen-google/pull/72
- knife-google - https://github.com/chef/knife-google/pull/159

tas50:
We should really make this a >= constraint TO avoid fallout in the community of consumers

skpaterson:
Thanks @tas50 they're updated to a >= constraint.  This has been successfully tested against the inspec-gcp resource pack but we should also wait until these are ready:
* kitchen-google - test-kitchen/kitchen-google#72
* knife-google - chef/knife-google#159
(^ I'll update these in line with this PR).

miah:
oops i have a extra space in there after fixing conflicts. will fix in subsequent PR.

Issue #288, opened 2018-04-25
Windows local transport doesn't provide command's stderr and exit code


grv87:
### Description
Windows' local transport command runner doesn't provide correct stderr and exit code.
See https://github.com/chef/inspec/issues/2993.

### Train and Platform Version
Train >= 0.31.0
Windows 7 SP1

### Replication Case
(with Inspec)
https://github.com/grv87/inspec_windows_command_stderr_exitcode

### Possible Solutions
I believe this regression was introduced in #220.
No problem with Train 0.30.0.

jerryaldrichiii:
I can confirm this is an issue with the local transport only.

```
You are currently running on:

    Name:      windows_server_2012_r2_standard_evaluation
    Families:  windows, os
    Release:   6.3.9600
    Arch:      x86_64
inspec> Inspec::VERSION
=> "2.2.35"
inspec> command('cmd /c exit 1').result
=> #<struct Train::Extras::CommandResult stdout="", stderr="", exit_status=0>
```

Will take some time in the coming days to investigate. Your hunch is right @grv87, this is due to how the local execution is changed in Train (keeping a session open and running commands vs spawning a new session for each command)

Also, this appears to only be an issue when using `cmd` inside of PowerShell. See:

```
You are currently running on:

    Name:      windows_server_2012_r2_standard_evaluation
    Families:  windows, os
    Release:   6.3.9600
    Arch:      x86_64
inspec> command('nope').exit_status
=> 1
```

There is also some funny business going on with stderr (when not using `cmd`) but I fixed that in: https://github.com/inspec/train/pull/320 which should be in InSpec soon.

jerryaldrichiii:
Given that I haven't done any work on this I'm going to remove my assignment.

Apologies if this surprises anyone.

PR #533, opened 2019-11-08
Return exit code for local Windows command

james-stocks:
Signed-off-by: James Stocks <jstocks@chef.io>

## Description
The exit code might not be 0 for a script that does not `throw` a PowerShell exception. We should get the actual exit code and return it.
Note that PowerShell has a `$?` which is boolean _as well as_ a numeric last exit code. The variable name train is using `exitstatus` is ambiguous on which it represents, maybe.

## Related Issue
Partly addresses #288 but that issue also requires the stderr to be provided.

## Types of changes

- [X ] Breaking change (a content change which would break existing functionality or processes)

If anyone is currently depending on the inaccurate 0 and 1 exit codes then they will get different behaviour when train starts returning the real exit code.

## Checklist:
- [X ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -200,6 +200,10 @@ module Train::Transports
               try {
                 $stdout = & $scriptBlock | Out-String
                 $exit_code = $LastExitCode
+                if ($exit_code -eq $null)
+                {
+                  $exit_code = 0
+                }
                 $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = $exit_code }
               } catch {
                 $stderr = $_ | Out-String
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -199,10 +199,12 @@ module Train::Transports
               $scriptBlock = $ExecutionContext.InvokeCommand.NewScriptBlock($command)
               try {
                 $stdout = & $scriptBlock | Out-String
-                $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = 0 }
+                $exit_code = $LastExitCode
+                $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = $exit_code }
               } catch {
                 $stderr = $_ | Out-String
-                $result = @{ 'stdout' = ''; 'stderr' = $stderr; 'exitstatus' = 1 }
+                $exit_code = $LastExitCode
+                $result = @{ 'stdout' = ''; 'stderr' = $stderr; 'exitstatus' = $exit_code }
               }
               $resultJSON = $result | ConvertTo-JSON
 
--- a/lib/train/transports/local.rb
+++ b/lib/train/transports/local.rb
@@ -146,6 +146,12 @@ module Train::Transports
           raise PipeError if @pipe.nil?
         end
 
+        # @param  cmd The command to execute
+        # @return Local::ComandResult with stdout, stderr and exitstatus
+        #         Note that exitstatus ($?) in PowerShell is boolean, but we use a numeric exit code.
+        #         A command that succeeds without setting an exit code will have exitstatus 0
+        #         A command that exits with an exit code will have that value as exitstatus
+        #         A command that fails (e.g. throws exception) before setting an exit code will have exitstatus 1
         def run_command(cmd)
           script = "$ProgressPreference='SilentlyContinue';" + cmd
           encoded_script = Base64.strict_encode64(script)
--- a/test/fixtures/PowerShell/exit_fortytwo.ps1
+++ b/test/fixtures/PowerShell/exit_fortytwo.ps1
@@ -1,0 +1,2 @@
+echo 'Goodbye'
+exit 42
--- a/test/fixtures/PowerShell/exit_zero.ps1
+++ b/test/fixtures/PowerShell/exit_zero.ps1
@@ -1,0 +1,1 @@
+echo 'Hello'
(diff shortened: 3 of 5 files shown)


lib/train/transports/local.rb
```diff
@@ -199,10 +199,12 @@ def start_pipe_server(pipe_name)
               $scriptBlock = $ExecutionContext.InvokeCommand.NewScriptBlock($command)
               try {
                 $stdout = & $scriptBlock | Out-String
-                $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = 0 }
+                $exit_code = $LastExitCode
+                $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = $exit_code }
               } catch {
```
james-stocks: I only need the exit code but at some point we should fix #288 fully and get stderr for passing scripts and stdout for failing scripts.
This whole section should maybe change because the `catch` block only applies to PowerShell scripts that `throw`. The catch block isn't entered for PowerShell scripts that "fail" by calling `exit n`

clintoncwolfe: For the change that's given here, this seems fine - you're not making the lack of STDOUT/STDERR worse.

stuartpreston: This looks reasonable as long as you are looking for the last exit code from the most recent Win32 process, and not the last PowerShell command was successful - [use `$?` for that](https://techibee.com/powershell/what-is-lastexitcode-and-in-powershell/1847)

james-stocks: I was taking the previous code's use of `0` to assume that it meant exit code and not exit status (in spite of the variable name)

james-stocks:
@miah @clintoncwolfe I added a test that mimics my own need (running a PS1 script that may or may not work).
There seems to be a bug where train has a problem with directly exiting a local Windows command.

e.g. if I attempt this test:

```
  it "run command with exit code" do
    cmd = conn.run_command('Write-Output "L8r allig8r"; Exit 42')
    _(cmd.stdout).must_equal "L8r allig8r\r\n"
    _(cmd.stderr).must_equal ""
    _(cmd.exit_status).must_equal 42
  end
```

it causes this error:

```
  2) Error:
windows local command#test_0003_run command with exit code:
EOFError: end of file reached
    C:/Users/jstocks/code/train/lib/train/transports/local.rb:154:in `readline'
    C:/Users/jstocks/code/train/lib/train/transports/local.rb:154:in `run_command'
    C:/Users/jstocks/code/train/lib/train/transports/local.rb:76:in `run_command_via_connection'
    C:/Users/jstocks/code/train/lib/train/plugins/base_connection.rb:128:in `run_command'
    test/windows/local_test.rb:40:in `block (2 levels) in <main>'

28 runs, 45 assertions, 2 failures, 1 errors, 0 skips
Traceback (most recent call last):
        3: from C:/Users/jstocks/code/train/lib/train/transports/local.rb:225:in `block in start_pipe_server'
        2: from C:/Users/jstocks/code/bundles/train/ruby/2.6.0/gems/win32-process-0.8.3/lib/win32/process.rb:797:in `kill'
        1: from C:/Users/jstocks/code/bundles/train/ruby/2.6.0/gems/win32-process-0.8.3/lib/win32/process.rb:797:in `each'
C:/Users/jstocks/code/bundles/train/ruby/2.6.0/gems/win32-process-0.8.3/lib/win32/process.rb:850:in `block in kill': Input/output error - TerminateProcess (Errno::EIO)
rake aborted!
```

I can put this failing test on a different branch and file an issue for it?

lib/train/transports/local.rb
```diff
@@ -199,10 +199,16 @@ def start_pipe_server(pipe_name)
               $scriptBlock = $ExecutionContext.InvokeCommand.NewScriptBlock($command)
               try {
                 $stdout = & $scriptBlock | Out-String
-                $result = @{ 'stdout' = $stdout ; 'stderr' = ''; 'exitstatus' = 0 }
+                $exit_code = $LastExitCode
+                if ($exit_code -eq $null)
+                {
+                  $exit_code = 0
```
james-stocks: This treats an unset `$LastExitCode` as `0`.
This is fiddling the result, but it preserves the previous behaviour where we set the exit_code to 0 if nothing went wrong.
I could _not_ do this, but it might break workflow if any current users are testing that 0 is returned for their command.

james-stocks:
@zenspider good call - I added 1 more unit test and a code comment to define the desired behaviour.
I can't easily find a good single reference article that explains PowerShell exit statuses.  PowerShell has both a boolean `$?` exit status variable for the success status of the last command and a numeric `$LASTEXITCODE` variable with the exit code value of the last time `exit` was called*; and they behave differently (but not totally independently).

* note potential confusing behaviour here - `$LASTEXITCODE` is $null for a new PowerShell session and it's value represents the last script that bothered to call `exit`, not the last script that executed.

To represent this properly, Windows and Linux might not be able to share a uniform `CommandResult` structure.

lib/train/transports/local.rb
```diff
@@ -146,6 +146,12 @@ def initialize(powershell_cmd = "powershell")
           raise PipeError if @pipe.nil?
         end
 
+        # @param  cmd The command to execute
+        # @return Local::ComandResult with stdout, stderr and exitstatus
+        #         Note that exitstatus ($?) in PowerShell is boolean, but we use a numeric exit code.
```
james-stocks: Code Climate says these comment lines are too long; but the Rakefile rubocop task doesn't mind them.
I'd rather have them this length for human readability. Do I need to shorten them?

miah: image

Codeclimate is not _required_ (yet).

Issue #529, opened 2019-11-01
opt-in cloud provider transport design


logicminds:
### Describe the Enhancement:
I want a more opt-in cloud provider transport design.

### Describe the Need:
The current train gem is too heavy weight pulling in aws, gce and azure cloud providers I don't use when I only want local docker support.  We should be able to just include the train gems we need instead of coupling them in with the main train gem. 

And the current train-core gem does not include docker. So there is no way to have a lightweight train-core and docker or aws, azure, gce.

All or nothing approach currently.

### Current Alternative
None. 

### Can We Help You Implement This?:
The new train-docker, train-aws, train-azure, train-gce need to be owned and published by inspec.

zenspider:
This can be the epic... but needs to be split out for each platform

PR #534, opened 2019-11-13
Adds activesupport gem to allow XML>JSON parsing from Azure APIs

r-fennell:
Signed-off-by: Ruairi Fennell <rfennell@chef.io>

## Description
Adds activesupport gem to allow XML>JSON parsing from Azure APIs.
Had originally added this as a direct dependency to inspec-azure but users have been running into problems as resource-pack dependencies don't appear to be installed/bundled. 

## Related Issue
https://github.com/inspec/inspec-azure/issues/216

## Types of changes

### DIFF
--- a/train.gemspec
+++ b/train.gemspec
@@ -29,6 +29,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "azure_mgmt_resources", "~> 0.15"
   spec.add_dependency "azure_graph_rbac", "~> 0.16"
   spec.add_dependency "azure_mgmt_key_vault", "~> 0.17"
+  spec.add_dependency "activesupport", "~> 5.2.3"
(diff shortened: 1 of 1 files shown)


miah:
Resource pack dependencies in Gemfiles are never installed. 

This is the same problem we had with inspec-aws. Dependencies will need to be added to train (or a train-<cloud>) project before being added to inspec-<cloud>.

r-fennell:
@kekaichinose Thanks. I agree the current situation around dependencies isn't ideal and i'm glad https://github.com/inspec/train/issues/529 has got the ball rolling on addressing some of the shortcoming in the train>inspec>resource-pack development & release cycle.
Of course feel free to loop me in to any further discussions you plan on having with @skpaterson .

Issue #538, opened 2019-11-24
Regression: Train::PlatformDetectionFailed: Sorry, we are unable to detect your platform


vsingh-msys:
# Version:
3.1.10

# Environment:
Chef Infra Client: 15.5.17

# Scenario:
Trying to bootstrap a ubuntu node which configured with `requiretty`. Seems the problem arises after https://github.com/inspec/train/commit/1eb18d490a0cf0a60cfb5754ee44d36bffabbf10 commit.

# Steps to Reproduce:

1. Spin a Ubuntu vagrant node and set requiretty by adding `Defaults requiretty` in the sudoers file.
2. Bootstrap the node

```
knife bootstrap node1.test -N node1.test -i .vagrant/machines/default/virtualbox/private_key -U vagrant --sudo --yes -VV
```

# Expected Result:
It should bootstrap the node

# Actual Result:
```
Train::PlatformDetectionFailed: Sorry, we are unable to detect your platform
 /train/lib/train/platforms/detect/scanner.rb:41:in `scan'
 /train/lib/train/platforms/detect.rb:9:in `scan'
 /train/lib/train/plugins/base_connection.rb:116:in `platform'
 /train/lib/train/extras/command_wrapper.rb:166:in `load'
 /train/lib/train/transports/ssh_connection.rb:53:in `initialize'
 /train/lib/train/transports/ssh.rb:240:in `new'
 /train/lib/train/transports/ssh.rb:240:in `create_new_connection'
 /train/lib/train/transports/ssh.rb:82:in `connection'
 /chef/lib/chef/knife/bootstrap/train_connector.rb:57:in `connection'
 /chef/lib/chef/knife/bootstrap/train_connector.rb:70:in `connect!'
 /chef/lib/chef/knife/bootstrap.rb:692:in `do_connect'
 /chef/lib/chef/knife/bootstrap.rb:620:in `connect!'
 /chef/lib/chef/knife/bootstrap.rb:569:in `run'
 /chef/lib/chef/knife.rb:486:in `block in run_with_pretty_exceptions'
 /chef/lib/chef/local_mode.rb:42:in `with_server_connectivity'
 /chef/lib/chef/knife.rb:485:in `run_with_pretty_exceptions'
 /chef/lib/chef/knife.rb:229:in `run'
 /chef/lib/chef/application/knife.rb:163:in `run'
 /chef/bin/knife:24:in `<top (required)>'
[...]

PR #539, opened 2019-11-26
Rolling back #515.

zenspider:
#515 causes regressions shown in #538 which I think is a bigger affected
set of users.

Until we can get this under test and reproducible, it needs to wait.

Signed-off-by: Ryan Davis <zenspider@chef.io>


zenspider:
cc/ @christian-wtd

zenspider:
This has been addressed via #544.

PR #549, opened 2019-12-20
Substitute require for require_relative

tas50:
require_relative is significantly faster and should be used when available. See @lamont-granquist for the full story.


zenspider:
Claims of speed should include or cite real benchmarks... Otherwise, this seems fine.

lamont-granquist:
in chef/chef I did that here:

https://github.com/chef/chef/pull/8508#issuecomment-490695435

PR #554, opened 2020-01-15
Refactor with_sudo_pty to BaseConnection (no-op) and SshConnection.

zenspider:
Fixes current red on master.


tas50:
This seems like a solid step in the right direction

PR #556, opened 2020-01-16
Kali Linux platform detection support

mattray:
Add support for detecting [Kali Linux](https://www.kali.org)

Signed-off-by: Matt Ray <github@mattray.dev>

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -76,6 +76,13 @@ module Train::Platforms::Detect::Specifications
             true
           end
         end
+      plat.name("kali").title("Kali Linux").in_family("debian")
+        .detect do
+          if linux_os_release && linux_os_release["ID"].eql?("kali")
+            @platform[:release] = linux_os_release["VERSION"]
+            true
+          end
+        end
       plat.name("raspbian").title("Raspbian Linux").in_family("debian")
         .detect do
           if (linux_os_release && linux_os_release["NAME"] =~ /raspbian/i) || \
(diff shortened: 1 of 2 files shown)


mattray:
FWIW I installed Chef via the .DEB on the Kali box and ohai works just fine. They will both report the same platform, family and version after this.

mattray:
Validated that this works with an aarch64 Kali install too

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -76,6 +76,13 @@ def self.load
             true
           end
         end
+      plat.name("kali").title("Kali Linux").in_family("debian")
+        .detect do
+          if linux_os_release && linux_os_release["ID"].eql?("kali")
+            @platform[:release] = linux_os_release["VERSION"]
```
zenspider: This is 3 calls to `linux_os_release`... please use local variables to make it only 1.

lib/train/platforms/detect/specifications/os.rb
```diff
@@ -76,6 +76,13 @@ def self.load
             true
           end
         end
+      plat.name("kali").title("Kali Linux").in_family("debian")
+        .detect do
+          if linux_os_release && linux_os_release["ID"].eql?("kali")
```
zenspider: `eql?` is _never_ the right comparator to call and can't succeed in the wild. Please stick with `==`.

test/unit/platforms/os_detect_test.rb
```diff
@@ -202,7 +216,7 @@ def debian_scan(id, version)
 
   describe "/etc/os-release" do
     describe "when not on a wrlinux build" do
-      it "fail back to genaric linux" do
+      it "fail back to generic linux" do
```
zenspider: Hah! I just did this too.

lib/train/platforms/detect/specifications/os.rb
mattray: Updated in latest PR.

lib/train/platforms/detect/specifications/os.rb
mattray: Updated in latest PR

PR #560, opened 2020-01-17
Fix Release detection on macOS

mattray:
This was returning '18.7.0' for the 10.14.6 Mojave release and it came
down to this was the version of the Darwin kernel instead of the macOS
release. The content of '/usr/bin/sw_vers' was being checked instead
of the output of the command, which was the source of the failure.

Signed-off-by: Matt Ray <github@mattray.dev>

### DIFF
--- a/lib/train/platforms/detect/specifications/os.rb
+++ b/lib/train/platforms/detect/specifications/os.rb
@@ -464,7 +464,7 @@ module Train::Platforms::Detect::Specifications
       plat.family("darwin").in_family("bsd")
         .detect do
           if unix_uname_s =~ /darwin/i
-            cmd = unix_file_contents("/usr/bin/sw_vers")
+            cmd = command_output("/usr/bin/sw_vers")
             unless cmd.nil?
               m = cmd.match(/^ProductVersion:\s+(.+)$/)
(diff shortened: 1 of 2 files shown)


mattray:
This now matches the output of ohai.

clintoncwolfe:
I think this is a very astute change, but I fear the consequences for some of our customers.

Consider existing profiles. If they are comparing against that kernel version, they are looking to see if it is (say) `>= 17`.  On my machine, prior to your change, that is true:

Screen Shot 2020-01-21 at 11 07 49

After this change, the meaning of that field changes, and the comparison suddenly fails on my machine (`"10.14.6" >= 17` is now false). That would introduce subtle errors, especially if users are using `platform` to conditionalize swaths of code in profiles, which is common.

I agree this would bring us more in line with ohai, but I think that has to be an all-at-once cut over (if that is the course taken) - we can't have some fields act the old way, and some fields act the new way; we'll need a major version cutover and have a "flag day" in which we say: hey, if you're running (say) InSpec 6, it acts like Ohai, so your profiles had better be 100% ready for that.

This is a great change, but it's just too early. I'd like to keep the branch but not accept the PR at this time - it is needed in the future.

PR #564, opened 2020-01-28
Train branch with AWS SDK 3.

bobchaos:

Implement AWS SDK v3

## Description
I'm not sure what was the author's intent, but presumably eleminate tech debt. I found this branch when poking around github issues for kitchen-terraform, so it helps with that.

## Types of changes

- [X] New content (non-breaking change)

## Checklist:

Existing test all pass locally, using `bundle exec rake test`

### DIFF
--- a/lib/train/transports/aws.rb
+++ b/lib/train/transports/aws.rb
@@ -1,0 +1,70 @@
+# encoding: utf-8
+require 'train/plugins'
+require 'aws-sdk-core'
+
+module Train::Transports
+  class Aws < Train.plugin(1)
+    name 'aws'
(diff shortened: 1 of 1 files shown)


bobchaos:
huh I must have done something very wrong, that diff should be 2 lines :/

bobchaos:
I tried again but same result, I'll try again when I've figured out what's happening

clintoncwolfe:
Hi @bobchaos , 
  Train has supported AWS SDK 3 for quite a while - over a year, I believe. Require the gem `train-aws` and you should be all set. AWS support is not included in Train core, and we're aiming to keep train core as slim as possible so that people can pick and choose what they want.

PR #566, opened 2020-01-31
Rebase #339

zenspider:
This rebases and rewrites #339 against current master. Mostly just style / testing issues were in conflict.


zenspider:
Since #339 was basically signed off on w/ a couple caveats, which are addressed here, then I think this is good to go.

PR #572, opened 2020-02-14
Revert to regular require to fix transport loading across gem boundary

clintoncwolfe:
Signed-off-by: Clinton Wolfe <clintoncwolfe@gmail.com>

Fixes https://github.com/inspec/inspec-azure/issues/234 . 

You can't require_relative a file when it is contained in a separate gem. So, our train-core / train split means we need to `require` these bundled files.

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/azure.rb
+++ b/lib/train/transports/azure.rb
@@ -1,15 +1,15 @@
 # encoding: utf-8
 
-require_relative "../plugins"
+require "train/plugins"
 require "ms_rest_azure"
 require "azure_mgmt_resources"
 require "azure_graph_rbac"
 require "azure_mgmt_key_vault"
 require "socket"
 require "timeout"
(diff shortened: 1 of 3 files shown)


lib/train/transports/azure.rb
```diff
@@ -1,15 +1,15 @@
...
+require "train/plugins"
 require "ms_rest_azure"
 require "azure_mgmt_resources"
 require "azure_graph_rbac"
 require "azure_mgmt_key_vault"
 require "socket"
 require "timeout"
-require_relative "helpers/azure/file_credentials"
-require_relative "clients/azure/graph_rbac"
-require_relative "clients/azure/vault"
```
btm: I don't _think_ you need these three because they're in the glob but :shipit:

PR #576, opened 2020-02-27
Attempt to fix --sudo.

skpaterson:

EDIT: this is a proposed fix for https://github.com/inspec/inspec/issues/4826

### DIFF
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -55,11 +55,10 @@ module Train::Extras
 
     # (see CommandWrapperBase::verify)
     def verify
+      # Do nothing, successfully. Don't use "true", that's not available on Windows.
       cmd = if @sudo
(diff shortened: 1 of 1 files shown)


tas50:
We'll want to test this with Chef Infra bootstrapping too since it's using the same code I believe.

clintoncwolfe:
I've performed manual testing against an ubuntu 16.04 vagrant box, and this worked well. 

I'm now going to test a freebsd11 machine.

clintoncwolfe:
After working with this a bit and discussing with @mp, we noted that running `run(-h)` (which would typically expand to `sh -c sudo -h` would only verify that you had sudo (or your alternate sudo command) installed and that it accepted the -h (likely usage help) option.  That's a semantic change from the previous `-v`, which would cause it to validate the credentials.

I've now tried it with `run(echo)` - basically "portably do nothing and succeed" on ubuntu16, freebsd11 on inspec, and both of those using knife bootstrap. That seems to work well.

PR #578, opened 2020-03-06
Avoid explicit /tmp in favor of $TMPDIR

majioa:
Signed-off-by: Malo Skrylevo <majioa@yandex.ru>

## Types of changes

## Checklist:

### DIFF
--- a/test/integration/bootstrap.sh
+++ b/test/integration/bootstrap.sh
@@ -1,22 +1,25 @@
 #!/bin/sh
-test ! -e /tmp/folder && \
-  mkdir /tmp/folder
-chmod 0567 /tmp/folder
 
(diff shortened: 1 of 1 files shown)


test/integration/bootstrap.sh
```diff
@@ -1,22 +1,22 @@
 #!/bin/sh
-test ! -e /tmp/folder && \
```
stevendanna: I wonder if we want to start this file with a quick check for the existence of `TMPDIR`, defaulting it to `/tmp/` if it doesn't otherwise exist.

majioa: Added changes

majioa:
@stevendanna thanx , please make patch release)

PR #583, opened 2020-04-08
[chef#9635] Add reason for sudo root password

vsingh-msys:
## Description

If the user has configured with sudo required root access
```
Defaults rootpw
user ALL=(ALL) PASSWD: ALL
```

It raises an error
```
"[sudo] password for root: Sorry, try again.\n[sudo] password for root: \nsudo: 1 incorrect password attempt"

``` 
Above need to catch and add `Train::UserError` sudo error.

## Related Issue

PR- https://github.com/chef/chef/pull/9635
Issues https://github.com/chef/chef/issues/9404, https://github.com/chef/chef/issues/2887

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -82,6 +82,8 @@ module Train::Extras
       when "sudo: sorry, you must have a tty to run sudo"
         ["Sudo requires a TTY. Please see the README on how to configure "\
           "sudo to allow for non-interactive usage.", :sudo_no_tty]
+      when /\[sudo\] password for root: Sorry, try again./
+        ["[sudo] password for root: Sorry, try again", :sudo_root_password_required]
       else
         [rawerr, nil]
       end
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -72,18 +72,16 @@ module Train::Extras
       rawerr = "#{res.stdout} #{res.stderr}".strip
 
       case rawerr
-      when "Sorry, try again"
+      when /Sorry, try again/
         ["Wrong sudo password.", :bad_sudo_password]
-      when "sudo: no tty present and no askpass program specified"
+      when /sudo: no tty present and no askpass program specified/
         ["Sudo requires a password, please configure it.", :sudo_password_required]
-      when "sudo: command not found"
+      when /sudo: command not found/
         ["Can't find sudo command. Please either install and "\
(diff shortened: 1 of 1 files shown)


lib/train/extras/command_wrapper.rb
```diff
@@ -82,6 +82,8 @@ def verify
       when "sudo: sorry, you must have a tty to run sudo"
         ["Sudo requires a TTY. Please see the README on how to configure "\
           "sudo to allow for non-interactive usage.", :sudo_no_tty]
+      when /\[sudo\] password for root: Sorry, try again./
+        ["[sudo] password for root: Sorry, try again", :sudo_root_password_required]
```
lamont-granquist, line 86: is there a need to distinguish between :sudo_root_password_required and :sudo_password_required and could this be made more generic by not matching the username explicitly in the prompt?

vsingh-msys: Yes, both are raised in different scenarios for example:

1. if --sudo option is provided and it has configured in a way to required a `Sudo password` to perform any action then it raises eg.
`sudo_password_required`

```
Defaults requiretty 

OR

Defaults rootpw

```

2. if --sudo & --use-sudo-option both are provided then, so in this case, we also set the given user password in `opts[:sudo_password]` but if 
given password is not for root use then it raises 

:sudo_root_password_required


Matching the username here(root) is because it only raises in condition if we have set `Defaults rootpw` in the sudoers file

lamont-granquist: Still feels like for train that this should be made generic and not match `root` specifically, and the label that comes back should not have "root" in it.  I see how `:sudo_password_required` is a very different error message, so it shouldn't reuse that label, but a better label should probably be used here.

vsingh-msys: So, Probably, we could have a better name for `:sudo_password_required` if it matches `sudo: no tty present and no askpass program specified` but it too late to decide

vsingh-msys: oh wait, we can match substring regex for `:bad_sudo_password`, as we attempt to login using the current user password

vsingh-msys:
```
case rawerr
when \Sorry, try again\
["Wrong sudo password.", :bad_sudo_password]
```

Is covering our purpose?

lamont-granquist: ah there we go, this makes much more sense now.

PR #586, opened 2020-05-08
Bump Chefstyle and ready for dependabot

tas50:
Remove the gemfile.local sourcing which dependabot can't work with. This
way we'll get nice PRs to bump deps including CVEs complete with
changelogs showing us what's changed.

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -54,7 +54,7 @@ class Train::Transports::SSH
       @bastion_user           = @options.delete(:bastion_user)
       @bastion_port           = @options.delete(:bastion_port)
 
-      @cmd_wrapper            = CommandWrapper.load(self, @transport_options)
+      @cmd_wrapper = CommandWrapper.load(self, @transport_options)
     end
 
     # (see Base::Connection#close)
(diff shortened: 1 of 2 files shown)


lib/train/transports/ssh_connection.rb
```diff
@@ -54,7 +54,7 @@ def initialize(options)
       @bastion_user           = @options.delete(:bastion_user)
       @bastion_port           = @options.delete(:bastion_port)
 
-      @cmd_wrapper            = CommandWrapper.load(self, @transport_options)
+      @cmd_wrapper = CommandWrapper.load(self, @transport_options)
```
Schwad, line 57: It looks relatively straightforward to me but I'll leave officially approval to someone on the core team who has had more time in train code than me. Quick question, was this a stylistic change?

james-stocks: It satisfies a cop in the current ChefStyle:

```
lib/train/transports/ssh_connection.rb:57:31: C: Layout/SpaceAroundOperators: Operator = should be surrounded by a single space.
```

Alternatively the indenting can be kept and the blank line at line 56 deleted - that also satisfies the cop.

This current form looks correct to me as this `@cmd_wrapper` isn't quite the same as the indented option passing lines above it.

PR #600, opened 2020-05-21
Ensure file --> exist works with Solaris10

### LINKED ISSUE #587: Remote scanning issue on Solaris 10 when using the sh shell
inspec --version
4.18.111

My test environment is Vagrant with a Solaris 10 image running on a Mac.
config.vm.box = "tnarik/solaris10-minimal"

My Customer is using a real Solaris 10 environment and sees the same issue.

Use the InSpec File resource in a remote scan to check for the presence of a file.

Create a simple control as below:
```
control "find_file_test_sol10-1" do
  impact 1.0
  describe command('test -e /etc/passwd') do
     its('exit_status') { should eq 0 }
  end
  describe file('/etc/passwd') do
    it{ should exist }
  end
end
```

Run `inspec exec <control> -t ssh://user@solarisbox`

The file is always found not to be there, but it does exist.

Debugging it was found that InSpec was running the `test -e` command to check if the file was there or not.
This was traced to train - https://github.com/inspec/train/blob/master/lib/train/file/remote/unix.rb

In this case remote scanning is done using the root account on the solaris 10 box, the default shell for the root account is /bin/sh and for this customer it is not possible to change that shell. Looking at the man pages you can see that the `-e` switch is not supported for the sh shell (https://docs.oracle.com/cd/E26505_01/html/816-5165/test-1.html#scrolltoc).

A workaround was to hack the train code to change the shell to bash:

```
def exist?

          @exist ||= begin

            f = @follow_symlink ? "" : " || test -L #{@spath}"

            @backend.run_command("env SHELL=/bin/bash test -e #{@spath}" + f)

              .exit_status == 0

          end

        end
```
But obviously they do not want to have to maintain that.

Another possibility may be to use the `-f` switch - but not sure of all the implications of that.

In looking to debug what was going on, I also looked at the file resource (https://github.com/inspec/inspec/blob/master/lib/inspec/resources/file.rb) which has a specific switch for Solaris in there as well.

Schwad:
# Description

See below issue for full context here. The short summary is that
Solaris10 does not support `test -e` flag in `sh`. This is a large
issue because 1. we support Solaris10 and 2. this is a silent failure.

The goal of this PR is to recognize a Solaris instance at the breaking
point, and then run the command with the provided `/usr/bin/test`
utility.

# Related Issues

Fixes #587

# Run on my machine

@lhasadreams provided fantastic guidance on the referenced issue for
replicating on your machine. In vagrant run the above control with
the vagrant box `tnarik/solaris10-minimal`:

```ruby
control "find_file_test_sol10-1" do
  impact 1.0
  describe command('test -e /etc/passwd') do
     its('exit_status') { should eq 0 }
  end
  describe file('/etc/passwd') do
    it{ should exist }
  end
end
```

(Note: make sure `sh` is your default shell in the `etc/passwd`)

On master both should fail. On my change second should pass.

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/file/remote/unix.rb
+++ b/lib/train/file/remote/unix.rb
@@ -22,8 +22,14 @@ module Train
         def exist?
           @exist ||= begin
             f = @follow_symlink ? "" : " || test -L #{@spath}"
-            @backend.run_command("test -e #{@spath}" + f)
-              .exit_status == 0
+            if @backend.platform.solaris?
+              # Solaris does not support `-e` flag in default `test`,
+              # so we specify by running /usr/bin/test:
+              # https://github.com/inspec/train/issues/587
+              @backend.run_command("/usr/bin/test -e #{@spath}" + f)
+            else
+              @backend.run_command("test -e #{@spath}" + f)
+            end.exit_status == 0
           end
         end
 


lib/train/file/remote/unix.rb
```diff
@@ -22,8 +22,13 @@ def content
         def exist?
           @exist ||= begin
             f = @follow_symlink ? "" : " || test -L #{@spath}"
-            @backend.run_command("test -e #{@spath}" + f)
-              .exit_status == 0
+            if @backend.run_command("uname -s").stdout.chomp == "SunOS"
+              # Solaris does not support `-e` flag:
```
james-stocks, line 26: This is confusing because we're still using -e flag. It should explain that the default `test` does not so we specify running `/usr/bin/test` instead

lib/train/file/remote/unix.rb
```diff
@@ -22,8 +22,13 @@ def content
         def exist?
           @exist ||= begin
             f = @follow_symlink ? "" : " || test -L #{@spath}"
-            @backend.run_command("test -e #{@spath}" + f)
-              .exit_status == 0
+            if @backend.run_command("uname -s").stdout.chomp == "SunOS"
```
james-stocks, line 25: I don't know train very well, but I wonder if the current target family/arch/version are already detected at this point and are accessible to this class.... that would save the round-trip on an additional command call every time.

lib/train/file/remote/unix.rb
```diff
@@ -22,8 +22,13 @@ def content
...
+              # Solaris does not support `-e` flag:
+              # https://github.com/inspec/train/issues/587
+              @backend.run_command("/usr/bin/test -e #{@spath}" + f)
+            else
+              @backend.run_command("test -e #{@spath}" + f)
+            end.exit_status == 0
```
james-stocks, line 31: I didn't know `end` yielded a value before, nice :)

lib/train/file/remote/unix.rb
Schwad: Really good point +1 thanks!

lib/train/file/remote/unix.rb
Schwad: This was the part that I spent the most time diving on I think. I hooked in to existing variables/methods, then looked around where you can inspect for version/etc but they weren't available here. I kept thinking *there must be a better way*. 

I'll try to pry back in and show off my instance vars and methods available here.

Schwad: Quick check with a pry:

```ruby
results_vars = {}
instance_variables.each do |k|
   results_vars[k] = eval k.to_s
end 

#=> {:@backend=>Train::Transports::SSH::Connection[unknown],
 :@path=>"/etc/passwd",
 :@follow_symlink=>true,
 :@spath=>"/etc/passwd",
 :@stat=>{:type=>:file, :mode=>420, :owner=>"root", :uid=>0, :group=>"sys", :gid=>3, :mtime=>1589990200, :size=>759, :selinux_label=>nil},
 :@mounted=>#<struct Train::Extras::CommandResult stdout="", stderr="", exit_status=1>,
 :@exist=>true,
 :@content=>
  "root:x:0:0:Super-User:/:/sbin/sh\ndaemon:x:1:1::/:\nbin:x:2:2::/usr/bin:\nsys:x:3:3::/:\nadm:x:4:4:Admin:/var/adm:\nlp:x:71:8:Line Printer Admin:/usr/spool/lp:\nuucp:x:5:5:uucp Admin:/usr/lib/uucp:\nnuucp:x:9:9:uucp Admin:/var/spool/uucppublic:/usr/lib/uucp/uucico\nsmmsp:x:25:25:SendMail Message Submission Program:/:\nlisten:x:37:4:Network Admin:/usr/net/nls:\ngdm:x:50:50:GDM Reserved UID:/:\nwebservd:x:80:80:WebServer Reserved UID:/:\npostgres:x:90:90:PostgreSQL Reserved UID:/:/usr/bin/pfksh\nsvctag:x:95:12:Service Tag UID:/:\nunknown:x:96:96:Unknown Remote UID:/:\nnobody:x:60001:60001:NFS Anonymous Access User:/:\nnoaccess:x:60002:60002:No Access User:/:\nnobody4:x:65534:65534:SunOS 4.x NFS Anonymous Access User:/:\nvagrant:x:100:100::/export/home/vagrant:/sbin/sh\n"}

results = {}
public_methods(false).each do |k|
   results[k] = send(k) rescue nil # in case args required
end  

#=> {:link_path=>nil,
 :path=>"/etc/passwd",
 :mounted=>#<struct Train::Extras::CommandResult stdout="", stderr="", exit_status=1>,
 :unix_mode_mask=>nil,
 :linked_to?=>nil,
 :content=>
  "root:x:0:0:Super-User:/:/sbin/sh\ndaemon:x:1:1::/:\nbin:x:2:2::/usr/bin:\nsys:x:3:3::/:\nadm:x:4:4:Admin:/var/adm:\nlp:x:71:8:Line Printer Admin:/usr/spool/lp:\nuucp:x:5:5:uucp Admin:/usr/lib/uucp:\nnuucp:x:9:9:uucp Admin:/var/spool/uucppublic:/usr/lib/uucp/uucico\nsmmsp:x:25:25:SendMail Message Submission Program:/:\nlisten:x:37:4:Network Admin:/usr/net/nls:\ngdm:x:50:50:GDM Reserved UID:/:\nwebservd:x:80:80:WebServer Reserved UID:/:\npostgres:x:90:90:PostgreSQL Reserved UID:/:/usr/bin/pfksh\nsvctag:x:95:12:Service Tag UID:/:\nunknown:x:96:96:Unknown Remote UID:/:\nnobody:x:60001:60001:NFS Anonymous Access User:/:\nnoaccess:x:60002:60002:No Access User:/:\nnobody4:x:65534:65534:SunOS 4.x NFS Anonymous Access User:/:\nvagrant:x:100:100::/export/home/vagrant:/sbin/sh\n",
 :mode=>420,
 :group=>"sys",
 :uid=>0,
 :gid=>3,
 :mode?=>nil,
 :sanitize_filename=>nil,
 :selinux_label=>nil,
 :size=>759,
 :mtime=>1589990200,
 :shallow_link_path=>nil,
 :type=>:file,
 :owner=>"root",
 :exist?=>true,
 :grouped_into?=>nil}
```

james-stocks: If `@backend` is available here then `@backend.platform` will have all the detected platform details

Schwad: Aha - yes. With a tiny look I found `@backend.platform.solaris?` which is *much* cleaner - thanks!

Issue #599, opened 2020-05-21
Problem with os_windows.rb when the target is cisco_ios


CodesWhisperer:
https://github.com/inspec/train/blob/c7ff3c059dc89fe68454b3c50b93c15b7a2d65db/lib/train/platforms/detect/helpers/os_windows.rb#L36-L38

When the target is a cisco_ios device, the exit_status is 0 but the error is described in the stdout, so the if condition doesn't work and the JSON.parse raises an exception.

To fix it, the if should be something like:
`return false if (command.exit_status != 0) || command.stdout.empty? || !(command.stdout.downcase =~ /window/)`

This line alone fixed it for me, but I also think the JSON.parse should be inside rescue block.

Schwad:
Hey @CodesWhisperer  👋 

Thanks for the issue! Would you like to open a PR with this change so we can continue the discussion there?

tas50:
Closing this out since we shipped the fix. Thanks @CodesWhisperer

PR #601, opened 2020-05-25
Check stdout even when exit_status is 0, and handle JSON.parse exception

CodesWhisperer:
## Description
When the target is a cisco_ios device, the exit_status is 0 but the error is described in the stdout, so the if condition doesn't work and the JSON.parse raises an exception.

Also, the JSON.parse should always be in rescue block.

## Related Issue
https://github.com/inspec/train/issues/599

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/platforms/detect/helpers/os_windows.rb
+++ b/lib/train/platforms/detect/helpers/os_windows.rb
@@ -37,7 +37,6 @@ module Train::Platforms::Detect::Helpers
 
       begin
         payload = JSON.parse(command.stdout)
-        
         @platform[:family] = "windows"
         @platform[:release] = payload["Version"]
         @platform[:name] = payload["Caption"]
@@ -45,7 +44,7 @@ module Train::Platforms::Detect::Helpers
         read_wmic
         true
       rescue
-        return false
+        false
       end
     end
 
--- a/lib/train/platforms/detect/helpers/os_windows.rb
+++ b/lib/train/platforms/detect/helpers/os_windows.rb
@@ -33,7 +33,8 @@ module Train::Platforms::Detect::Helpers
       command = @backend.run_command(
         "Get-WmiObject Win32_OperatingSystem | Select Caption,Version | ConvertTo-Json"
       )
-      return false if (command.exit_status != 0) || command.stdout.empty? || !(command.stdout.downcase =~ /window/)
+      # some targets (e.g. Cisco) may return 0 and print an error to stdout
+      return false if (command.exit_status != 0) || command.stdout.downcase !~ /window/
 
       begin
         payload = JSON.parse(command.stdout)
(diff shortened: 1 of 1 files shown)


clintoncwolfe:
Code looks reasonable to me, just need to get it past the linter. Also please sign your commit to indicate acceptance of our contribution terms - https://github.com/inspec/train/blob/master/CONTRIBUTING.md#developer-certification-of-origin-dco

CodesWhisperer:
@clintoncwolfe 
I fixed it, but the tests "tests-ruby-2.6-windows" didn't pass and I don't know why.

clintoncwolfe:
The gem ffi segfaults under windows right now. We have a pending PR to pin the version back to a working version. Once that merges you can rebase to master and the windows tests should pass.

lib/train/platforms/detect/helpers/os_windows.rb
```diff
@@ -33,15 +33,19 @@ def check_powershell
...
-      return false if (command.exit_status != 0) || command.stdout.empty?
-
-      payload = JSON.parse(command.stdout)
-      @platform[:family] = "windows"
-      @platform[:release] = payload["Version"]
-      @platform[:name] = payload["Caption"]
-
-      read_wmic
-      true
+      return false if (command.exit_status != 0) || command.stdout.empty? || !(command.stdout.downcase =~ /window/)
```
james-stocks, line 36: `!(command.stdout.downcase =~ /window/)` could be slightly simpler `command.stdout.downcase !~ /window/`

This new condition seems to make `command.stdout.empty?` redundant so the line could be

`return false if (command.exit_status != 0)  || command.stdout.downcase !~ /window/`

It may be valuable to leave a comment line above this statement to explain this case (some targets e.g. Cisco may return 0 and print an error to stdout)

james-stocks:
Hi @CodesWhisperer -

This looks good, thank you! I left one comment suggesting to make the code easier to understand; because in future an issue may occur with some other platform; and the logic might get changed in a way that fixes that other platform but causes Cisco to start failing again...

As Clinton said, please rebase this branch against master to get the Windows tests working.

Thanks

tas50:
@CodesWhisperer Can you rebase this against master please?

rmoles:
It would be good to get this merged. Could you rebase against master @CodesWhisperer ?

rmoles:
@CodesWhisperer unfortunately your last commit is missing your DCO, could you please push another commit with your DCO added?

CodesWhisperer:
@rmoles Is it good?

rmoles:
Awesome thanks @CodesWhisperer. Unfortunately, I don't have merge rights on this repo. Perhaps, @james-stocks or @Schwad can merge this.

james-stocks (approved):

james-stocks:
Thanks for this @CodesWhisperer !

Issue #612, opened 2020-06-17
SSH Connection Reuse is Broken


tecracer-theinen:
# Version:

3.3.1

# Environment:

Ubuntu 18.04

# Scenario:

When executing multiple commands over the SSH transport, the session is closed and reopened for every single one. This results in greatly longer execution times, especially as platform detection kicks off each time.

Reason is a wrong comparison between current and requested connection options in [ssh.rb line 79](https://github.com/tecracer-theinen/train/blob/master/lib/train/transports/ssh.rb#L79).

# Steps to Reproduce:

Create Connection with debug logging, execute multiple commands.

# Expected Result:

```
[2020-06-17T16:27:13+00:00] DEBUG: [SSH] opening connection to ubuntu@1.2.3.4
...
[2020-06-17T16:27:19+00:00] DEBUG: [SSH] ubuntu@1.2.3.4 cmd = env lsmod
[2020-06-17T16:27:19+00:00] DEBUG: [SSH] reusing existing connection ubuntu@1.2.3.4
[2020-06-17T16:27:19+00:00] DEBUG: [SSH] ubuntu@1.2.3.4 cmd = lsb_release -a
[2020-06-17T16:27:19+00:00] DEBUG: [SSH] reusing existing connection ubuntu@1.2.3.4
[2020-06-17T16:27:19+00:00] DEBUG: [SSH] reusing existing connection ubuntu@1.2.3.4
```

# Actual Result:

```
[2020-06-17T16:27:13+00:00] DEBUG: [SSH] opening connection to ubuntu@1.2.3.4
...
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] ubuntu@1.2.3.4 cmd = env lsmod
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] shutting previous connection ubuntu@1.2.3.4
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] closing connection to ubuntu@1.2.3.4
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] opening connection to ubuntu@1.2.3.4
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] ubuntu@1.2.3.4 cmd = cmd.exe /c ver
[2020-06-17T16:09:41+00:00] DEBUG: [SSH] ubuntu@1.2.3.4 cmd = Get-WmiObject Win32_OperatingSystem | Select Caption,Version | ConvertTo-Json
[...]

PR #613, opened 2020-06-17
Fix SSH Connection reuse

tecracer-theinen:
## Description

This PR fixes the comparison between current and requested SSH connection options.

## Related Issue
Fixes issue #612 

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -65,6 +65,9 @@ module Train::Transports
     option :non_interactive, default: false
     option :verify_host_key, default: false
 
+    # Allow connecting with older algorithms
+    option :append_all_supported_algorithms, default: true
+
     option :compression_level do |opts|
       # on nil or false: set compression level to 0
       opts[:compression] ? 6 : 0
@@ -76,7 +79,7 @@ module Train::Transports
       validate_options(opts)
       conn_opts = connection_options(opts)
 
-      if defined?(@connection) && @connection_options == conn_opts
+      if defined?(@connection) && reusable_connection?(conn_opts)
         reuse_connection(&block)
       else
         create_new_connection(conn_opts, &block)
@@ -84,6 +87,13 @@ module Train::Transports
     end
 
     private
+
+    def reusable_connection?(conn_opts)
+      return false unless @connection_options
+
+      # Do requested options match their current settings
+      @connection_options.all? { |k, v| conn_opts[k] == v }
+    end
 
     def validate_options(options)
       super(options)
@@ -167,6 +177,7 @@ module Train::Transports
         bastion_user: opts[:bastion_user],
         bastion_port: opts[:bastion_port],
         non_interactive: opts[:non_interactive],
+        append_all_supported_algorithms: opts[:append_all_supported_algorithms],
         transport_options: opts,
       }
       # disable host key verification. The hash key and value to use
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -202,8 +202,6 @@ class Train::Transports::SSH
         require "net/ssh/proxy/command"
         @options[:proxy] = Net::SSH::Proxy::Command.new(generate_proxy_command)
       end
-      # Allow older algorithms
-      @options[:append_all_supported_algorithms] = true
       Net::SSH.start(@hostname, @username, @options.clone.delete_if { |_key, value| value.nil? })
     rescue *RESCUE_EXCEPTIONS_ON_ESTABLISH => e
       if (opts[:retries] -= 1) <= 0


lib/train/transports/ssh.rb
```diff
@@ -76,7 +76,7 @@ def connection(state = {}, &block)
       validate_options(opts)
       conn_opts = connection_options(opts)
 
-      if defined?(@connection) && @connection_options == conn_opts
+      if reusable_connection?(conn_opts)
```
james-stocks, line 79: Why remove the `defined?(@connection)` test? I don't foresee the case where `@connection` would be `nil` but it seems wise to guard against the possibility of it.

lib/train/transports/ssh.rb
```diff
@@ -85,6 +85,15 @@ def connection(state = {}, &block)
 
     private
 
+    def reusable_connection?(conn_opts)
+      return false unless @connection_options
+
+      current_options = @connection_options
+      current_options.reject! { |k| %i{append_all_supported_algorithms}.include? k }
```
james-stocks, line 92: `append_all_supported_algorithms` is needed in the ssh options to allow connections to use older and less secure algorithms. Is there a reason it gets discarded here?

james-stocks:
I'd like @tas50 to get a look at this in case it affects how chef or knife depends on train.

lib/train/transports/ssh.rb
tecracer-theinen: I will have a look at this one. But when I worked on this, it appeared that that option was removed somewhere on the current connection options. I'll come back to this

lib/train/transports/ssh.rb
tecracer-theinen: Good point. Overambitious optimization on my side :)

lib/train/transports/ssh.rb
james-stocks: (if it's only intended to be excluded from the comparison between the two option hashes, then line 91 should maybe use `current_options = @connection_options.dup` to take a copy of the options instead of a reference to `@connection_options`)

james-stocks: Isn't the bug that the newly added `append_all_supported_algorithms` option is missing from the connection_options hash at https://github.com/tecracer-theinen/train/blob/master/lib/train/transports/ssh.rb#L145 ?

james-stocks:
@tecracer-theinen we need https://github.com/inspec/train/commit/a3a73a8adaa9dd6f3205a04efb32d70155a80ce9 to put the `append_all_supported_algorithms` option in the right place.
But even after doing that, we still need this PR to get working re-use.

I'm not sure how the logic `if defined?(@connection) && @connection_options == conn_opts` ever worked because when I test this, `@connection_options` includes all options for the run, not just the ssh opts...

```
{:logger=>#<Logger:0x00007ff254a535e0 @level=4, @progname=nil, @default_formatter=#<Logger::Formatter:0x00007ff254a53568 @datetime_format=nil>, @formatter=nil, @logdev=#<Logger::LogDevice:0x00007ff254a53518 @shift_period_suffix=nil, @shift_size=nil, @shift_age=nil, @filename=nil, @dev=#<IO:<STDOUT>>, @binmode=false, @mon_data=#<Monitor:0x00007ff254a534c8>, @mon_data_owner_object_id=6800>>, :user_known_hosts_file=>"/dev/null", :hostname=>"172.30.1.5", :port=>22, :username=>"vagrant", :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>15, :connection_retries=>5, :connection_retry_sleep=>1, :max_wait_until_ready=>600, :auth_methods=>["none", "publickey"], :keys_only=>true, :keys=>["/Users/jstocks/inspec_test/.vagrant/machines/default/virtualbox/private_key"], :password=>nil, :forward_agent=>nil, :proxy_command=>nil, :bastion_host=>nil, :bastion_user=>"root", :bastion_port=>22, :non_interactive=>false, :append_all_supported_algorithms=>true, :transport_options=>{:host=>"172.30.1.5", :port=>22, :user=>"vagrant", :key_files=>["/Users/jstocks/inspec_test/.vagrant/machines/default/virtualbox/private_key"], :shell=>false, :shell_options=>nil, :shell_command=>nil, :sudo=>false, :sudo_options=>nil, :sudo_password=>nil, :sudo_command=>nil, :password=>nil, :keepalive=>true, :keepalive_interval=>60, :connection_timeout=>15, :connection_retries=>5, :connection_retry_sleep=>1, :max_wait_until_ready=>600, :compression=>false, :pty=>false, :proxy_command=>nil, :bastion_host=>nil, :bastion_user=>"root", :bastion_port=>22, :non_interactive=>false, :verify_host_key=>false, :append_all_supported_algorithms=>true, :compression_level=>0, :auth_methods=>["none", "publickey"], :keys_only=>true}, :verify_host_key=>false}
```

So we do need your change from this PR that instead tests that all keys in the ssh options match keys in the current options.

james-stocks:
The branch https://github.com/inspec/train/commits/js/fix_option now contains 1 commit to fix the `append_all_supported_algorithms` option and a second commit to copy the fixed options comparison from this PR.

@tecracer-theinen I can make that branch a PR, or we can adapt this existing PR, which do you prefer.

And thank you for helping fix this!

lib/train/transports/ssh.rb
```diff
@@ -85,6 +88,14 @@ def connection(state = {}, &block)
 
     private
 
+    def reusable_connection?(conn_opts)
+      return false unless @connection_options
+
+      current_options = @connection_options.dup
```
phiggins, line 94: If `@connection_options` is not mutated, why is a `.dup` call necessary here?

lib/train/transports/ssh.rb
```diff
@@ -85,6 +88,14 @@ def connection(state = {}, &block)
...
+
+      current_options.all? { |k, v| conn_opts[k] == v }
```
phiggins, line 96: Drive-by comment: this looks like a bug to me. If the intended behavior is to test that `@connection_options` is a subset of `conn_opts` and not that they are equal, a comment would be a nice way to indicate that to future code readers.

lib/train/transports/ssh.rb
tecracer-theinen: Just followed the suggestion here, assuming there was some Ruby knowledge beyond me. If it's not needed, I am happy to drop it.

james-stocks: I asked for it on an earlier commit where we were calling `reject!` on `current_options` - it can be removed now.

Issue #615, opened 2020-06-25
Invalid error Train::PlatformDetectionFailed when AIX password has expired.


vsingh-msys:
# Version:
3.3.3

# Environment:

- knife bootstrap AIX 7.2 machine

# Scenario:
- Bootstrap a machine `bundle exec knife bootstrap IP_ADDRESS -N aix-7.2 -U user --sudo --yes -VV
`
- Found AIX machine password has expired and by running 

```
> @backend.run_command(cmd)
DEBUG: [SSH] vsingh@172.31.10.28 cmd = uname -s
#<struct Train::Extras::CommandResult stdout="", stderr="WARNING: Your password has expired.\nPassword change required but no TTY available.\n", exit_status=255>
```

# Expected Result:
It should be bootstrap the AIX 7.2 machine.

# Actual Result:

```
bundle exec knife bootstrap IP_ADDRESS -N aix-7.2 -U user --sudo --yes -VV
INFO: Using configuration from /chef-workstation/.chef/config.rb
DEBUG: Checking if we need to accept Chef license to bootstrap node
DEBUG: Reading products and relationships...
DEBUG: Successfully read products and relationships
DEBUG: License acceptance required for chef version: 16
DEBUG: Searching for the following licenses: ["infra-client", "inspec"]
DEBUG: Found license chef_infra_client at /etc/chef/accepted_licenses/chef_infra_client
DEBUG: Found license inspec at /etc/chef/accepted_licenses/inspec
DEBUG: Missing licenses remaining: []
DEBUG: All licenses present
Connecting to IP_ADDRESS using ssh
DEBUG: [SSH] opening connection to user@IP_ADDRESS
DEBUG: [SSH] using options {:user_known_hosts_file=>"/dev/null", :port=>22, :compression=>false, :compression_level=>0, :keepalive=>true, :keepalive_interval=>60, :timeout=>60, :auth_methods=>["none", "password", "keyboard-interactive"], :keys_only=>nil, :keys=>[], :password=>"<hidden>", :forward_agent=>nil, :non_interactive=>true, :verify_host_key=>:always}
Connecting to IP_ADDRESS using ssh
DEBUG: [SSH] opening connection to user@IP_ADDRESS
[...]

PR #616, opened 2020-06-25
Fix incorrect error message when password is expired

vsingh-msys:
## Description

I have come across a scenario where the AIX system password got expired due to this getting platform detention error which is wrong. Fixes provide an exact error message to the user so that, action should be taken accordingly.

## Related Issue

Fixes: https://github.com/inspec/train/issues/615

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/platforms/detect/helpers/os_common.rb
+++ b/lib/train/platforms/detect/helpers/os_common.rb
@@ -35,17 +35,19 @@ module Train::Platforms::Detect::Helpers
 
     def command_output(cmd)
       res = @backend.run_command(cmd)
-      output = res.stdout
-      output = res.stderr if output == "" && res.exit_status != 0
       # When you try to execute command using ssh connction as root user and you have provided ssh user identity file
       # it gives standard output to login as authorised user other than root. To show this standard ouput as an error
       # to user we are matching the string of stdout and raising the error here so that user gets exact information.
-      if @backend.class.to_s == "Train::Transports::SSH::Connection" && (output =~ /Please login as the user/ || output =~ /WARNING: Your password has expired.\nPassword change required but no TTY available.\n/)
-        raise Train::UserError, "SSH failed: #{output}"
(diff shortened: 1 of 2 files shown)


lib/train/platforms/detect/helpers/os_common.rb
```diff
@@ -34,16 +34,18 @@ def unix_file_exist?(path)
...
+      res = @backend.run_command(cmd)
+      output = res.stdout
+      output = res.stderr if output == ""  && res.exit_status != 0
       # When you try to execute command using ssh connction as root user and you have provided ssh user identity file
       # it gives standard output to login as authorised user other than root. To show this standard ouput as an error
       # to user we are matching the string of stdout and raising the error here so that user gets exact information.
-      if @backend.class.to_s == "Train::Transports::SSH::Connection" && res =~ /Please login as the user/
-        raise Train::UserError, "SSH failed: #{res}"
+      if @backend.class.to_s == "Train::Transports::SSH::Connection" && (output =~ /Please login as the user/ || output =~ /WARNING: Your password has expired.\nPassword change required but no TTY available.\n/)
+        raise Train::UserError, "SSH failed: #{output}"
```
vsingh-msys, line 44: this might not be an optimal solution, any suggestion would be appreciated. Thanks!

james-stocks: Maybe this would be more readable as 

```
if @backend.class.to_s == "Train::Transports::SSH::Connection" && (tdout =~ /Please login as the user/ 
    raise Train::UserError, "SSH failed: #{stdout}"
end
if @backend.class.to_s == "Train::Transports::SSH::Connection" && stderr =~ /WARNING: Your password has expired/)
    raise Train::UserError, "SSH failed: #{stderr}"
end
```

There's duplication between the two `if` statements but I think it's OK because it's expressing two separate failure cases.

PR #625, opened 2020-07-14
Allow timeout option per ssh command

james-stocks:
The option is per-command not per-connection so that end users can timeout individual InSpec tests.
The default is no timeout - timing out has the side effect of leaving commands running on the host. It should only be used where the end user is OK with this.

Signed-off-by: James Stocks <jstocks@chef.io>

### DIFF
--- a/lib/train/errors.rb
+++ b/lib/train/errors.rb
@@ -40,4 +40,7 @@ module Train
 
   # Exception for when a invalid cache type is passed.
   class UnknownCacheType < Error; end
+
+  # Exception for when a command reaches configured timeout
+  class CommandTimeoutReached < Error; end
 end
--- a/lib/train/plugins/base_connection.rb
+++ b/lib/train/plugins/base_connection.rb
@@ -127,17 +127,30 @@ class Train::Plugins::Transport
     # This is the main command call for all connections. This will call the private
     # run_command_via_connection on the connection with optional caching
     #
-    # Supported options for the opts hash:
-    # :timeout    Number of seconds to allow before a timeout. Implemented in the
-    #             derived connection class (currently ssh_connection)
-    #
     # This command accepts an optional data handler block. When provided,
     # inbound data will be published vi `data_handler.call(data)`. This can allow
     # callers to receive and render updates from remote command execution.
+    #
+    # @param [String] the command to run
+    # @param [Hash] optional hash of options for this command. The derived connection
+    #               class's implementation of run_command_via_connection should receive
+    #               and apply these options.
     def run_command(cmd, opts = {}, &data_handler)
-      return run_command_via_connection(cmd, opts, &data_handler) unless cache_enabled?(:command)
+      # Some implementations do not accept an opts argument.
+      # We cannot update all implementations to accept opts due to them being separate plugins.
+      # Therefore here we check the implementation's arity to maintain compatibility.
+      case method(:run_command_via_connection).arity.abs
+      when 1
+        return run_command_via_connection(cmd, &data_handler) unless cache_enabled?(:command)
 
-      @cache[:command][cmd] ||= run_command_via_connection(cmd, opts, &data_handler)
+        @cache[:command][cmd] ||= run_command_via_connection(cmd, &data_handler)
+      when 2
+        return run_command_via_connection(cmd, opts, &data_handler) unless cache_enabled?(:command)
+
+        @cache[:command][cmd] ||= run_command_via_connection(cmd, opts, &data_handler)
+      else
+        raise NotImplementedError, "#{self.class} does not implement run_command_via_connection with arity #{method(:run_command_via_connection).arity}"
+      end
     end
 
     # This is the main file call for all connections. This will call the private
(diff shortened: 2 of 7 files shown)


james-stocks:
@clintoncwolfe already pointed out that `Timeout` is bad](https://medium.com/@adamhooper/in-ruby-dont-use-timeout-77d9d4e5a001) so I don't think this approach is good idea :(

james-stocks:
This might have to be ssh_connection specific then an equivalent change made to `inspec-winrm` - there may be a way to neatly(-ish) terminate the command e.g. https://github.com/net-ssh/net-ssh/blob/master/lib/net/ssh/connection/channel.rb#L331

james-stocks:
The PR as of https://github.com/inspec/train/pull/625/commits/ac09315062b14453df4700b3f906fc03a6b62691 works for basic testing. This needs tested very carefully as it introduces threading.

lib/train/transports/ssh_connection.rb
```diff
@@ -334,7 +336,20 @@ def execute_on_channel(cmd)
           end
         end
       end
-      session.loop
+
+      thr = Thread.new { session.loop }
+
```
Schwad, line 341: I ran some of this in a few test files and after a think I do believe this is a proper and safe use of Ruby's `Thread`.

Much of the threadsafety issues I would worry about have to do with asynchronous interruptions, creating state (especially if the state is passed around) and managing multiple threads. 

Since this new code keeps it to the one thread with no fancy state moving in/out/around the thread and its contents, that should be good. 

As I understand it `#join` pauses any further code execution until the contents of the thread block either finish or the limit (timeout) is reached. In that case, I believe we wouldn't have to worry about other Threads getting jammed on the back of this one. 

I am happy to swap this PR to an approval any time but will leave it currently as a comment as I know we've requested a review from Foundations to ensure this plays well with the rest of the ecosystem. :)

lib/train/plugins/base_connection.rb
```diff
@@ -134,12 +134,19 @@ def platform
...
     # callers to receive and render updates from remote command execution.
-    def run_command(cmd, opts = {}, &data_handler)
+    def run_command(cmd, opts, &data_handler)
       return run_command_via_connection(cmd, opts, &data_handler) unless cache_enabled?(:command)
 
       @cache[:command][cmd] ||= run_command_via_connection(cmd, opts, &data_handler)
     end
 
+    # For backward compatibility with plugins, a form of run_command without opts
+    def run_command(cmd, &data_handler)
```
james-stocks, line 144: This isn't Java, you can't do this :)

james-stocks:
This now passes CI, but the code to handle different method arity's needs to:
1. have new unit tests added
2. maybe update https://github.com/inspec/train/blob/master/docs/plugins.md#run_command_via_connection

lib/train/transports/ssh_connection.rb
james-stocks: This PR is almost entirely different from when you first looked at it; so taking another look before approving would be appreciated, thanks!

As we were seeing earlier, this change needs to work well for a. a subsequent change to `train-winrm` and b. all our current plugin users

lib/train/errors.rb
```diff
@@ -40,4 +40,7 @@ class PlatformDetectionFailed < Error; end
 
   # Exception for when a invalid cache type is passed.
   class UnknownCacheType < Error; end
+
+  # Exception for when a command reaches configured timeout
```
james-stocks, line 44: This _maybe_ should not be here, since it's only used by ssh_connection, and that's technically a plugin. I don't feel it's wrong to have a set of errors in core train that are standard for plugins though.

Schwad:
Note to self: Having another quick look over and with Clinton's eyes on this, I am happy to approve.

However I am holding fire until stocksy gives his final okay as I don't want this to accidentally get merged without his final thumbsup.

lib/train/errors.rb
james-stocks: I definitely want to implement this for winrm as well

lib/train/plugins/base_connection.rb
```diff
@@ -130,10 +130,27 @@ def platform
...
+    #
+    # @param [String] the command to run
+    # @param [Hash] optional hash of options for this command. The derived connection
+    #               class's implementation of run_command_via_connection should receive
+    #               and apply these options.
+    def run_command(cmd, opts = {}, &data_handler)
+      # Some implementations do not accept an opts argument.
+      # We cannot update all implementations to accept opts due to them being separate plugins.
+      # Therefore here we check the implementation's arity to maintain compatibility.
+      case method(:run_command_via_connection).arity.abs
```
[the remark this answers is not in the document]
james-stocks: Thanks @clintoncwolfe :) . One possible improvement here is to log a warning if the implementation does not support an _opts_ arg but some opts were given.

james-stocks:
Thanks very much @clintoncwolfe @Schwad 
I need to complete some final testing - I was in the middle of this before heading out on leave.

james-stocks:
@clintoncwolfe  @Schwad I've tested this out with a combination of debian, RHEL and SUSE targets using both Vagrant and cloud VMs. 
Looks good to me. If there are any edge cases, note that this timeout needs to be set via option, and execution is largely unchanged for existing users of train.

clintoncwolfe (approved):

PR #629, opened 2020-08-10
Add new Azure resource pack backend dependencies

omerdemirok:
Signed-off-by: Omer Demirok <odemirok@chef.io>

## Description
The new Azure resource pack [backend](https://github.com/inspec/inspec-azure/tree/azure_generic_resources) requires following additional dependencies to be added:
- [Dotenv](https://github.com/bkeepers/dotenv)
- [ipaddr](https://github.com/ruby/ipaddr)
- [facets](https://github.com/rubyworks/facets)
## Types of changes

## Checklist:

### DIFF
--- a/train.gemspec
+++ b/train.gemspec
@@ -41,4 +41,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "docker-api", "~> 1.26"
   spec.add_dependency "google-api-client", ">= 0.23.9", "< 0.35.0"
   spec.add_dependency "googleauth", ">= 0.6.6", "< 0.11.0"
+  spec.add_dependency "dotenv", "~> 2.7.6"
+  spec.add_dependency "facets", "~> 3.1.0"
+  spec.add_dependency "ipaddr", "~> 1.2"
 end
--- a/train.gemspec
+++ b/train.gemspec
@@ -38,10 +38,8 @@ Gem::Specification.new do |spec|
   spec.add_dependency "azure_mgmt_resources", "~> 0.15"
   spec.add_dependency "azure_mgmt_security", "~> 0.18"
   spec.add_dependency "azure_mgmt_storage", "~> 0.18"
+  spec.add_dependency "ipaddr", "~> 1.2" # Azure, for IP address comparison.
   spec.add_dependency "docker-api", "~> 1.26"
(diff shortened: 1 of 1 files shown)


omerdemirok:
CC: @rmoles, @skpaterson

train.gemspec
```diff
@@ -41,4 +41,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "docker-api", "~> 1.26"
   spec.add_dependency "google-api-client", ">= 0.23.9", "< 0.35.0"
   spec.add_dependency "googleauth", ">= 0.6.6", "< 0.11.0"
+  spec.add_dependency "dotenv", "~> 2.7.6"
+  spec.add_dependency "facets", "~> 3.1.0"
+  spec.add_dependency "ipaddr", "~> 1.2"
```
clintoncwolfe, line 46: A small ask: maybe a comment noting that these are required for azure, or group them into their own section using a comment and some whitespace?

train.gemspec
```diff
@@ -41,4 +41,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency "docker-api", "~> 1.26"
   spec.add_dependency "google-api-client", ">= 0.23.9", "< 0.35.0"
   spec.add_dependency "googleauth", ">= 0.6.6", "< 0.11.0"
+  spec.add_dependency "dotenv", "~> 2.7.6"
+  spec.add_dependency "facets", "~> 3.1.0"
```
tas50, line 45: Is this something we really need or are we just pulling these in so we can use some helpers? Train gets installed in a lot of places and this doesn't seem like a must have for implemented azure support.

omerdemirok: Working on removing `facets` and `dotenv`. However, `ipaddr` is essential to make security rules comparison work properly. It is provided in the AWS resource pack.

omerdemirok:
Found out that [`ipaddr`](https://github.com/ruby/ipaddr#installation) is shipped in Ruby distros. 
There is no need to add a new dependency at this point.
Thanks, @clintoncwolfe, @tas50, and @Schwad.

Schwad:
Thanks for your work in this area @omerdemirok ! Much appreciated 🙏

tas50:
I'm glad it turned out we didn't need this one. Keeping this dep list as small as possible is an important goal.

Issue #633, opened 2020-09-04
su: must be run from a terminal


vsingh-msys:
# Version:
3.3.6 or latest master

# Environment:
Vagrant Ubuntu 18.04

# Scenario:

In order to perform knife bootstrap with a different root password to the authenticated user https://github.com/chef/chef/issues/10181
it required to hit `su - USER` command

# Steps to Reproduce:

```
require 'train'
train = Train.create('ssh', host: 'node1.test', port: 22, user: 'vagrant', key_files: './.vagrant/machines/default/virtualbox/private_key')

conn = train.connection
res = conn.run_command('su - root')

puts "STDOUT: #{res.stdout}"
puts "STDERR: #{res.stderr}"

>
STDOUT:
STDERR: su: must be run from a terminal

```

# Expected Result:

it should at least open the new session for suing user.

# Actual Result:

Raises an error `STDERR: su: must be run from a terminal`

PR #636, opened 2020-09-10
su - USER command execution support & change transport_options to attr_accessor

vsingh-msys:
## Description

 - Modify transport options whenever required.
 - su - USER command execution using -c require a password ie. `su - USER -c 'ls;date;'`
 - Append the `su_password` option if command execution required a password.

## Related Issue

Fixes: https://github.com/inspec/train/issues/633

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -29,8 +29,8 @@ class Train::Transports::SSH
   #
   # @author Fletcher Nichol <fnichol@nichol.ca>
   class Connection < BaseConnection # rubocop:disable Metrics/ClassLength
-    attr_reader :hostname
-    attr_reader :transport_options
+    attr_reader   :hostname
+    attr_accessor :transport_options
 
     def initialize(options)
       # Track IOS command retries to prevent infinite loop on IOError. This must
@@ -316,8 +316,12 @@ class Train::Transports::SSH
         channel.exec(cmd) do |_, success|
           abort "Couldn't execute command on SSH." unless success
           channel.on_data do |_, data|
-            yield(data) if block_given?
-            stdout += data
+            if data == "Password: "
+              channel.send_data "#{@transport_options[:su_password]}\n"
+            else
+              yield(data) if block_given?
+              stdout += data
+            end
           end
 
           channel.on_extended_data do |_, _type, data|
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -315,7 +315,7 @@ class Train::Transports::SSH
 
         channel.exec(cmd) do |_, success|
           abort "Couldn't execute command on SSH." unless success
-          channel.on_data do |ch, data|
+          channel.on_data do |_, data|
             yield(data, channel) if block_given?
             stdout += data
           end
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -242,6 +242,16 @@ class Train::Transports::SSH
 
       exit_status, stdout, stderr = execute_on_channel(cmd, &data_handler)
 
+      # An interactive console might contain the STDERR in STDOUT
+      # concat both outputs for non-zero exit status. 
+      output = "#{stdout} #{stderr}".strip if exit_status != 0
+
+      # Abstract the su - USER authentication failure
+      # raise the Train::UserError and passes message & reason
+      if output && output.match?("su: Authentication failure")
+        raise Train::UserError.new(output, :bad_su_user_password)
+      end
+
       # Since `@session.loop` succeeded, reset the IOS command retry counter
       @ios_cmd_retries = 0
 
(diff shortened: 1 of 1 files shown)


lib/train/transports/ssh_connection.rb
```diff
@@ -316,8 +316,12 @@ def execute_on_channel(cmd)
         channel.exec(cmd) do |_, success|
           abort "Couldn't execute command on SSH." unless success
           channel.on_data do |_, data|
-            yield(data) if block_given?
-            stdout += data
+            if data == "Password: "
+              channel.send_data "#{@transport_options[:su_password]}\n"
```
lamont-granquist, line 320: This should use the accessor.  Also it looks like any time the data contains "Password: " the su_password will be output, even if it is not set?  It also isn't clear to me why this isn't just done
in the caller by passing a block that looks for "Password: " and outputs the su_password, and why this needs to go into train.

lib/train/transports/ssh_connection.rb
```diff
@@ -316,12 +316,12 @@ def execute_on_channel(cmd)
         channel.exec(cmd) do |_, success|
           abort "Couldn't execute command on SSH." unless success
           channel.on_data do |_, data|
-            yield(data) if block_given?
+            yield(data, channel) if block_given?
```
vsingh-msys, line 319: Now we can consume this on the train side with password the data block
```
require 'train'
train = Train.create('ssh', host: 'node1.test', port: 22, user: 'vagrant', key_files: './.vagrant/machines/default/virtualbox/private_key', pty: true)
conn = train.connection

password = "PASSWORD"
res = conn.run_command("su - root -c 'date;'") do |data, ch|
	ch.send_data("#{password}\n") if data.match?("Password:")
end

puts "STDOUT: #{res.stdout}"
puts "STDERR: #{res.stderr}"
```

lib/train/transports/ssh_connection.rb
```diff
@@ -316,12 +316,12 @@ def execute_on_channel(cmd)
...
             stdout += data
           end
 
           channel.on_extended_data do |_, _type, data|
-            yield(data) if block_given?
+            yield(data, channel) if block_given?
```
vsingh-msys, line 324: Added to maintain the parity with `on_data` block.

lib/train/transports/ssh_connection.rb
```diff
@@ -242,6 +242,16 @@ def run_command_via_connection(cmd, &data_handler)
...
 
+      # An interactive console might contain the STDERR in STDOUT
+      # concat both outputs for non-zero exit status. 
+      output = "#{stdout} #{stderr}".strip if exit_status != 0
+
+      # Abstract the su - USER authentication failure
+      # raise the Train::UserError and passes message & reason
+      if output && output.match?("su: Authentication failure")
+        raise Train::UserError.new(output, :bad_su_user_password)
+      end
```
vsingh-msys, line 253: Wrap the `su: Authentication failure` run_command error message and raise custom `Train::UserError` with reason `bad_su_user_password`

vsingh-msys:
@clintoncwolfe could you please have a look? Thanks!

tas50:
@clintoncwolfe Any chance you can chime in on this change?

lib/train/transports/ssh_connection.rb
clintoncwolfe: I've tested using the given code and it does work. I also verified that the new block parameter is optional, so any existing code that calls run_command without the `channel` param will not break.

This does put all of the responsibility for handling the password on the calling code.

lib/train/transports/ssh_connection.rb
```diff
@@ -29,8 +29,8 @@ class Train::Transports::SSH
   #
   # @author Fletcher Nichol <fnichol@nichol.ca>
   class Connection < BaseConnection # rubocop:disable Metrics/ClassLength
-    attr_reader :hostname
-    attr_reader :transport_options
+    attr_reader   :hostname
+    attr_accessor :transport_options
```
clintoncwolfe, line 33: Could you help me understand why this change is needed? I don't see anywhere in the PR what the transport_option ivar is directly changed, and if you need to change one of its elements you can do so without making it read-write.

lamont-granquist: Here's the context where the goal is to change the pty option to true later in the process:

https://github.com/chef/chef/pull/10410/files#r492238172

I had the same question, but didn't know enough about train to suggest anything better.

james-stocks: Does this work OK with the ssh code [here](https://github.com/inspec/train/blob/master/lib/train/transports/ssh.rb#L81) that looks at the options to determine if the connection can be re-used or needs restarted?

vsingh-msys: @james-stocks As we memoinzed the connection object at chef core side https://github.com/chef/chef/blob/0a29892953b07ba1fded119f5aef364e14c7dd07/lib/chef/knife/bootstrap/train_connector.rb#L51-L59, won't be a problem here, but not sure about other impacts.

PR #637, opened 2020-09-11
Allow for docker-api 2.x

tas50:
docker-api resolves warnings with Ruby 2.7 and removes the pin to legacy docker API releases which kept it from working properly on Windows platforms.


Schwad (approved):
Thanks!

PR #638, opened 2020-09-11
Switch from __FILE__ to __dir__ where we can

tas50:
This simplifies dir determination


stevendanna (approved):
Nice

PR #642, opened 2020-09-25
Support unpacking train options from URI query values

lamont-granquist:
Pretty simple to support arbitrary options in URIs so that stuff like `ssh://user:pass@host.com:123/path?bastion_host=stronghold.example.org&bastion_user=bulwark` becomes valid.  This came up as a result of the work in https://github.com/chef/ohai/pull/1503 to support ohai over train and to allow passing a URL into ohai with the `--target` option on the command line such that it doesn't require the presence of a `.chef/crendentials` file.

### DIFF
--- a/lib/train.rb
+++ b/lib/train.rb
@@ -113,6 +113,9 @@ module Train
     # TODO: rewrite next line using compact! once we drop support for ruby 2.3
     creds = creds.delete_if { |_, value| value.nil? }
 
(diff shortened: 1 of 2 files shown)


lamont-granquist:
paging @clintoncwolfe and @Schwad for review

clintoncwolfe (approved):

PR #649, opened 2020-12-07
Make :forward_agent a valid SSH option

drbrain:
## Description

When I run `knife bootstrap --forward-agent …` chef calls Train::options to fetch the valid options for an SSH connection and rejects any options that are not listed as valid for Train::Transport::SSH:

https://github.com/chef/chef/blob/v16.8.2/lib/chef/knife/bootstrap/train_connector.rb#L269-L280

One of the options validated is `:forward_agent`.  Since it is not listed as a valid option by the SSH transport it can never be set on any connection.

Without this change Train::Transport::SSH connections for `knife bootstap` will always include `-o ForwardAgent=no` regardless of the user's choices.

Previously :forward_agent was not listed as a valid option for the SSH transport.  Now it is included as a valid option.

This means that `knife bootstrap --forward-agent` will now generate an SSH command containing `-o ForwardAgent=yes`.

The option is false by default in order to match current train behavior.

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -63,6 +63,7 @@ module Train::Transports
     option :bastion_port, default: 22
     option :non_interactive, default: false
     option :verify_host_key, default: false
+    option :forward_agent, default: false
 
     # Allow connecting with older algorithms
     option :append_all_supported_algorithms, default: true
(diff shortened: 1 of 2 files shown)


lamont-granquist (approved):

tas50:
Thanks for the fix @drbrain. This is up on rubygems now and building in Chef Workstation as well. I opened a ticket to make sure we fully test this scenario on the client side as well.

PR #659, opened 2021-01-20
Read the username and port from /.ssh/config file and replace if present

### LINKED ISSUE #101: Read port from ssh config file (e.g. /home/user/.ssh/config)
### Description

Read port also from the ssh config under /home/user.ssh/config.
### InSpec and Platform Version

inspec version is 0.20.1 and archlinux 4.5.2-1-ARCH
### Replication Case

/home/user/.ssh/config look like this

Host jumphost
        Hostname 1.1.1.1
        Port 28646

inspec exec ./ -t ssh://root@jumphost --key-files /home/user/.ssh/id_rsa is not working because of the port. I have to add the Port also

inspec exec ./ -t ssh://root@jumphost --key-files /home/user/.ssh/id_rsa --port 28646

I think train read the config partly, because i can use the configured hostnames in /home/user/.ssh/config.

sanga1794:
Signed-off-by: sanga17 <sausekar@msystechnologies.com>

## Description
Read the username and port from /.ssh/config file and replace if present

## Related Issue
Fixes #101

## Types of changes

## Checklist:


lib/train/transports/ssh_connection.rb
```diff
@@ -52,6 +52,7 @@ def initialize(options)
       @bastion_host           = @options.delete(:bastion_host)
       @bastion_user           = @options.delete(:bastion_user)
       @bastion_port           = @options.delete(:bastion_port)
+      check_config_options
```
clintoncwolfe, line 55: I might suggest the name `read_options_from_ssh_config`. The word `check` implies that there is some validation going on.

clintoncwolfe, line 55: I think this needs to happen first, not last. If a user provides a username (or port) in the connection options as well as in the SSH config files, we want the provided option to take precedence. As written, the SSH config file will overwrite the user-provided option, which is not what we want.

clintoncwolfe (changes requested):
I think this is a great contribution! However, we need to make sure that the precedence is correct. Thanks!

lib/train/transports/ssh_connection.rb
sanga1794: sure will do @clintoncwolfe, thanks

sanga1794: Hi @clintoncwolfe, i have tried earlier to use it at first, but its not working as we are expecting, when we are not providing the user, SSH config file is not overwriting it, as we are setting default user value to 'root' https://github.com/inspec/train/blob/ef35cec62de6f0ace9361e2d2fe73169d7878332/lib/train/transports/ssh.rb#L47

sanga1794:
@clintoncwolfe please have a look on updated PR. Thanks!

lib/train/transports/ssh_connection.rb
clintoncwolfe: We can't have that precedence - it needs to be CLI options, config file, default. Adding your change will make it config file, CLI options, default. Consider adding code above the call to `super(options)` to detect if port and username are being set and if so don't override them in read_options_from_ssh_config.

clintoncwolfe:
This PR is waiting on an update from @sanga1794 to fix the precedence. Do you think you will have time to make the change?

sanga1794:
@clintoncwolfe I have updated the precedence, please have a look. Thanks!!

lib/train/options.rb
```diff
@@ -59,6 +59,9 @@ def merge_options(base, opts)
           default = hm[:default]
           if default.is_a? Proc
             res[field] = default.call(res)
+          elsif hm.key?(:coerce)
+            field_value = hm[:coerce].call(res[:host])
```
clintoncwolfe, line 63: I don't think it makes sense in general to pass the hostname to the Proc. Instead, just pass the options hash (`res`)  to the Proc, and let the Proc internally access the field that it needs (if any - here both will need host).

clintoncwolfe (commented):
Overall this looks good! I just have one minor change requested - to alter what is passed to the Proc.

lib/train/options.rb
sanga1794: Sure i'll make those changes. Thanks @clintoncwolfe !!

clintoncwolfe (approved):
OK! Thanks for your work on this - I think it has a lot of future potential!

Issue #358, opened 2018-09-28
Train leftover PowerShell process on local transport


jquick:
Train leftover PowerShell process

Issue Summary: Orphaned PowerShell processes are stacking up after being spawned by Inspec/train for audit runs on Windows Server 2012 R2

Expected Results: powershell.exe killed appropriately

Actual Results: powershell.exe processes hanging around

Repro Steps (if possible): I have not been able to reproduce

Request for Developers: Debug assistance

Customer scanning with Audit Cookbook and compliance profile "DevSec Windows Security Baseline". When he decodes the parameter for the logic being run from one of the orphaned PowerShell processes, it's this portion: https://github.com/inspec/train/blob/master/lib/train/transports/local.rb#L188-L217

I've matched the environment as closely as possible, and I have not yet been able to reproduce the issue. Does engineering have any ideas as to what might be causing this, or suggestions on other troubleshooting methods?

Vasu1105:
We have identified the same issue while debugging the multiple Powershell Process are getting created while running Windows Server 2019 remediation cookbook.  The reason behind this is we create the conn using local transport `conn = Train.create('local').connection` it creates the runner pipe here https://github.com/inspec/train/blob/master/lib/train/transports/local.rb#L26, https://github.com/inspec/train/blob/e03fcb161d28cf648dbe6fdbb199556be9d97234/lib/train/transports/local.rb#L51,
https://github.com/inspec/train/blob/e03fcb161d28cf648dbe6fdbb199556be9d97234/lib/train/transports/local.rb#L144,
and this method gets called https://github.com/inspec/train/blob/e03fcb161d28cf648dbe6fdbb199556be9d97234/lib/train/transports/local.rb#L185
in which it this Powershell Process is getting created https://github.com/inspec/train/blob/e03fcb161d28cf648dbe6fdbb199556be9d97234/lib/train/transports/local.rb#L229-L231 but that gets killed only when the Train execution is complete which is nothing but the ruby execution https://github.com/inspec/train/blob/e03fcb161d28cf648dbe6fdbb199556be9d97234/lib/train/transports/local.rb#L234

We also found that there is no implementation of close method where all the processes or sessions which are getting created should get killed. 

We explicitly called the Process.kill with server_pid which resulted into closing that Powershell process.

Vasu1105:
Also like to tag this issue in this as it also talks about the same #475

Vasu1105:
[...]

PR #660, opened 2021-01-25
Powershell process keep running with local train transport.

Vasu1105:

## Description

While running compliance remediation cookbook for Windows Server 2019 one of the customer has reported that its creating multiple Powershell processes as defined in issue #358 after debugging the same.

This PR adds the implementation of close method which kills all running processes once the execution of the command is done.
Issue can be reproduce with following
```
cmds = ["dir", "mkdir test", "hostname", "$env:PATH"]
cmds.each do |cmd|
  conn = Train.create('local').connection
  conn = conn.run_command(cmd)
 end
```
Testing close method

```
cmds = ["dir", "mkdir test", "hostname", "$env:PATH"]
cmds.each do |cmd|
  conn = Train.create('local').connection
  conn = conn.run_command(cmd)
  conn.close
 end
```

## Types of changes

## Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,24 +1,27 @@
-<!-- latest_release -->
-<!-- latest_release -->
-
-<!-- release_rollup -->
-<!-- release_rollup -->
-
-<!-- latest_stable_release -->
+<!-- latest_release 3.4.7 -->
 ## [v3.4.7](https://github.com/inspec/train/tree/v3.4.7) (2021-01-11)
 
 #### Merged Pull Requests
-- Update chefstyle requirement from 1.5.7 to 1.5.9 [#654](https://github.com/inspec/train/pull/654) ([dependabot-preview[bot]](https://github.com/dependabot-preview[bot]))
-- Update google API upper constraints to allow Ruby 3 [#656](https://github.com/inspec/train/pull/656) ([clintoncwolfe](https://github.com/clintoncwolfe))
 - Test on ruby 3.0, drop testing on ruby 2.4 [#657](https://github.com/inspec/train/pull/657) ([clintoncwolfe](https://github.com/clintoncwolfe))
-<!-- latest_stable_release -->
+<!-- latest_release -->
 
+<!-- release_rollup since=3.4.4 -->
+### Changes not yet released to rubygems.org
+
+#### Merged Pull Requests
+- Test on ruby 3.0, drop testing on ruby 2.4 [#657](https://github.com/inspec/train/pull/657) ([clintoncwolfe](https://github.com/clintoncwolfe)) <!-- 3.4.7 -->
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,27 +1,24 @@
-<!-- latest_release 3.4.7 -->
-## [v3.4.7](https://github.com/inspec/train/tree/v3.4.7) (2021-01-11)
-
-#### Merged Pull Requests
-- Test on ruby 3.0, drop testing on ruby 2.4 [#657](https://github.com/inspec/train/pull/657) ([clintoncwolfe](https://github.com/clintoncwolfe))
+<!-- latest_release -->
```
clintoncwolfe, line 1: Changes to the CHANGELOG should happen automaticaly when the PR merges - including them on a PR will only introduce conflicts.  These changes should not be included.

lib/train/transports/local.rb
```diff
@@ -160,12 +174,22 @@ def run_command(cmd)
...
         def acquire_pipe
+          require "win32/process"
+
           pipe_name = "inspec_#{SecureRandom.hex}"
 
-          start_pipe_server(pipe_name)
+          @server_pid = start_pipe_server(pipe_name)
+
+          # Ensure process is killed when the Train process exits
+          at_exit { Process.kill("KILL", @server_pid) unless @server_pid.nil? }
```
mwrock, line 192: my ruby is rusty but could you call `self.close` instead of repeating the `Process.kill` here

mwrock (approved):

CHANGELOG.md
Vasu1105: Definitely this got pushed by mistake will revert this one. Thanks @clintoncwolfe

lib/train/transports/local.rb
Vasu1105: Thanks @mattray.  Did the required changes.

tecracer-theinen: I hit that exact line of code yesterday for another reason. Maybe we can include the fix to that here as well?

When programmatically using InSpec from a custom Chef resource, the broken pipe of the KILL signal results in a failed Chef run as the `at_exit` will make the run exit non-zero on exceptions. I had a simple `rescue true` suffix for the `Process.kill` command at hand.

Question only remains if this PR should be merged first and I hand in my addition of if we append the fix here.

clintoncwolfe:
It looks like the added test is failing the linter - run `bundle exec chefstyle -a` to fix.

lib/train/transports/local.rb
clintoncwolfe: At this point, we're only killing the subprocess if we know the PID - that is, we believe that the process is running. So I think that should cover the common cases. I think absorbing all exceptions will likely hide new problems. I'm reluctant to add a catch-all rescue unless we must.

clintoncwolfe (approved):
Thanks!

PR #664, opened 2021-02-11
Implement upload/download for connections

tecracer-theinen:
Adds default `upload`/`download` methods to transfer files. Also adds a `content=` method for remote file manipulation.

## Description

While the SSH and the WinRM transport plugin already implemented the `upload`/`download` methods, those were not part of the general plugin API.

Offering these two methods and general file manipulation capabilities greatly extends any tool built on Train (e.g. Chef Target Mode).

On `base_connection`, the `upload` method iterates using the new `content=` method which uses base64 en-/decoding on Unix/Windows platforms. The `local` transport got simple local file manipulations for this, the `cisco_ios_connection`will throw an error. Can't test QNX/AIX, sorry.

I actually prepared both a Test Kitchen "train" transport and a remote shell utility based on this and it's working fine with my Linux/Windows systems. So I suppose the same goes for any other transport (AWS Systems Manager, VMware GOM etc).

## Related Issue

none

## Types of changes

- [X] New content (non-breaking change)

## Checklist:

### DIFF
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -24,7 +24,7 @@ module Train
     # interface methods: these fields should be implemented by every
     # backend File
     DATA_FIELDS = %w{
-      exist? mode owner group uid gid content content= mtime size selinux_label path
+      exist? mode owner group uid gid content mtime size selinux_label path
     }.freeze
 
     DATA_FIELDS.each do |m|
--- a/lib/train/file.rb
+++ b/lib/train/file.rb
@@ -24,7 +24,7 @@ module Train
     # interface methods: these fields should be implemented by every
     # backend File
     DATA_FIELDS = %w{
-      exist? mode owner group uid gid content mtime size selinux_label path
+      exist? mode owner group uid gid content content= mtime size selinux_label path
     }.freeze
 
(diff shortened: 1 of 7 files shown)


tecracer-theinen:
Definitively need help with the `content=` test in `mock.rb` - I don't understand that one :sweat_smile:

lib/train/file.rb
```diff
@@ -24,7 +24,7 @@ def sanitize_filename(_path)
     # interface methods: these fields should be implemented by every
     # backend File
     DATA_FIELDS = %w{
-      exist? mode owner group uid gid content mtime size selinux_label path
+      exist? mode owner group uid gid content content= mtime size selinux_label path
```
clintoncwolfe, line 27: I found that if you rename the method to be `set_content`, you can avoid the name issue, but then you encounter arity issues in that all of the other data fields are expected to be called with no arguments. If the new method is omitted from the DATA_FIELDS list - it is not really a data field, it is a setter - and instead treated as an abstract method, the tests pass.

tecracer-theinen: That was an easy fix. Thank you!

PR #676, opened 2021-04-13
Fix SSH Timeout PTY allocation

clintoncwolfe:
## Description

Do not automatically allocate a PTY when a timeout is given for SSH connections - it changes the semantics of commands such as `ls`.

## Related Issue

https://github.com/inspec/inspec/issues/5459
## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -323,7 +323,7 @@ class Train::Transports::SSH
 
         logger.debug("[SSH] #{self} cmd = #{cmd}")
 
-        if @transport_options[:pty] || timeout
+        if @transport_options[:pty]
           channel.request_pty do |_ch, success|
             raise Train::Transports::SSHPTYFailed, "Requesting PTY failed" unless success
           end


tas50:
We've had a lot of TTY changes made recently for knife bootstrap, one of which broke many bootstrap scenarios. Before we merge this we need additional testing to make sure it doesn't impact the boostrap scenarios.

clintoncwolfe:
After discussion with @tas50, I explained that this change only affects situations using the timeout feature of the SSH transports, which is relatively new and has been broken since being introduced. The change is to revert back to previous behavior, only allocating a PTY when explicitly requested. He felt comfortable with the merge if I am confident in low impact of the change, which I am.

james-stocks:
FYI @clintoncwolfe the reason for the PTY being allocated when using a timeout was to ensure that the command being executed would terminate on the target host. 

e.g. if you run `command("sleep 5000", timeout: 5)` , when train times out after 5 seconds you do not want to see the `sleep` command still running on the host. Without this PTY, if I remember correctly, you will see the sleep process still waiting the remining 4995 seconds.
If a command runs indefinitely, then processes will stack up as InSpec does repeated compliance scans and repeatedly runs the command then gives up on it.

Seems like forcing a PTY isn't the right way to do that...

PR #679, opened 2021-04-20
Switch to GNU timeout-based implementation of SSH timeouts

### LINKED ISSUE #677: SSH Timeout Feature Leaves Behind Processes on Target Host
This is a formalization of the problem @james-stocks brought up [here](https://github.com/inspec/train/pull/676#issuecomment-819523899)

# Version:

train 3.6.2
inspec 4.32.0

# Environment:

Seen on CentOS 8, but any SSH target would likely exhibit the same

# Scenario:

Trying to us the command timeout feature of inspec over a train SSH transport. A timed out command appears to time out and returns control to the inspec shell, but in fact the process lingers on the target host.

# Steps to Reproduce:

Run CentOS 8 in a Vagrant machine
In one terminal, connect to the VM and run `top -u vagrant` and watch for processes running `sleep`
Connect using inspec shell, I used InSpec 4.32.0
```
inspec shell -b ssh --host 127.0.0.1 -i /Users/cwolfe/sandbox/inspec/vagrants/centos-8/.vagrant/machines/default/virtualbox/private_key --user vagrant --port 2222
Welcome to the interactive InSpec Shell
To find out how to use it, type: help

You are currently running on:

    Name:      centos
    Families:  redhat, linux, unix, os
    Release:   8.3.2011
    Arch:      x86_64

inspec> command("sleep 60", timeout: 2).stdout
Inspec::Exceptions::ResourceFailed: Command `sleep 60` timed out after 2 seconds
from /Users/cwolfe/sandbox/inspec/inspec-dev/lib/inspec/resources/command.rb:64:in `rescue in result'
Caused by Train::CommandTimeoutReached: ssh command reached timeout (2s)
from /Users/cwolfe/sandbox/inspec/train/lib/train/transports/ssh_connection.rb:365:in `execute_on_channel'
```
# Expected Result:

Note that the timeout happens within 2-3 seconds, but the `top` terminal shows the `sleep` process running the full 60 seconds.

# Proposed solutions

The critical code is [here](https://github.com/inspec/train/blob/68887148ad7b0e25c59fe6f3a9167f55de0001d6/lib/train/transports/ssh_connection.rb#L356).

Considered, with notes:

 * Allocating a PTY. This causes the processes to exit, but breaks input semantics for many commands such as `ls` and breaks many profiles. 
 * Sending Control-C down the channel. Control-C is interpreted by the terminal into an INT signal sent to the last PID. No terminal, no interpretation; also we don't know the PID or PIDs.
 * Wrapping all calls in the GNU `timeout` CLI utility instead of relying on threading. This relies on the utility being present, and will also interfere with command wrapping / piping (which is very common  with `--sudo`)
 * Calling [`session.close()`](https://net-ssh.github.io/net-ssh/Net/SSH/Connection/Session.html) . Attempted, but it politely blocks.
 * Calling `session.shutdown!` Attempted, but it does not kill the process.
 * Calling `session.exec!("pkill -INT -fx #{cmd}"` in the timeout handler. This is a scary thing to do, since cmd is opaque to us. This does work in simple cases, but does not work with --sudo or other situations with pipes.
 * Limiting the blast radius by making the --command-timeout option in inspec have no default, so only people who opt-in will run the risk of abandoned processes.
 * Rolling back the timeout feature from the SSH transport
 * Rolling back the entire command timeout feature

Other ideas are certainly welcome.

clintoncwolfe:
## Description

As the title says, this PR switches the SSH timeout feature to use the GNU timeout utility (when available) to implement command timeouts instead of using a client-side thread and closing the session, which abandons processes on the server (See #677). 

This severely limits the applicability of the feature. First, the timeout utility must be installed. As it is part of gnu coreutils, it is common on Linux and other systems, but not universal, so we must do a check for it. Second, as it is a command wrapper, it does not function properly with any command string that includes pipes. Notably, all InSpec sudo commands include pipes. So, we check for pipes, and do not attempt to do a timeout on any command containing pipes. 

This PR retains some functionality while making the feature safe to use. It is a compromise between letting the processes continue to run as on #677, and withdrawing the feature entirely from the SSH transport.

## Related Issue

Fixes #677

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh_connection.rb
+++ b/lib/train/transports/ssh_connection.rb
@@ -321,6 +321,9 @@ class Train::Transports::SSH
         # wrap commands if that is configured
         cmd = @cmd_wrapper.run(cmd) if @cmd_wrapper
 
+        # Timeout the command if requested and able
+        cmd = "timeout #{timeout}s #{cmd}" if timeout && timeoutable?(cmd)
+
         logger.debug("[SSH] #{self} cmd = #{cmd}")
 
         if @transport_options[:pty]
@@ -350,21 +353,30 @@ class Train::Transports::SSH
           end
         end
       end
+      session.loop
 
-      thr = Thread.new { session.loop }
-
-      if timeout
-        res = thr.join(timeout)
-        unless res
-          logger.debug("train ssh command '#{cmd}' reached requested timeout (#{timeout}s)")
-          session.channels.each_value { |c| c.eof!; c.close }
-          raise Train::CommandTimeoutReached.new "ssh command reached timeout (#{timeout}s)"
-        end
-      else
-        thr.join
+      if timeout && timeoutable?(cmd) && exit_status == 124
+        logger.debug("train ssh command '#{cmd}' reached requested timeout (#{timeout}s)")
+        session.channels.each_value { |c| c.eof!; c.close }
+        raise Train::CommandTimeoutReached.new "ssh command reached timeout (#{timeout}s)"
       end
 
       [exit_status, stdout, stderr]
+    end
+
+    # Returns true if we think we can attempt to timeout the command
+    def timeoutable?(cmd)
+      have_timeout_cli? && !cmd.include?("|") # Don't try to timeout a command that has pipes
+    end
+
+    # Returns true if the GNU timeout command is available
+    def have_timeout_cli?
+      return @have_timeout_cli unless @have_timeout_cli.nil?
+
+      res = session.exec!("timeout --version")
+      @have_timeout_cli = res.exitstatus == 0
+      logger.debug("train ssh have_timeout_cli status is '#{@have_timeout_cli}'")
+      @have_timeout_cli
     end
   end
 end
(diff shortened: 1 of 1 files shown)


Vasu1105:
Thanks @clintoncwolfe this looks good to me. May be having test cases around it will be good, if possible.

Vasu1105 (approved):

lib/train/transports/ssh_connection.rb
```diff
@@ -350,21 +353,30 @@ def execute_on_channel(cmd, opts)
...
-      if timeout
-        res = thr.join(timeout)
-        unless res
-          logger.debug("train ssh command '#{cmd}' reached requested timeout (#{timeout}s)")
-          session.channels.each_value { |c| c.eof!; c.close }
-          raise Train::CommandTimeoutReached.new "ssh command reached timeout (#{timeout}s)"
-        end
-      else
-        thr.join
+      if timeout && timeoutable?(cmd) && exit_status == 124
```
Nik08, line 358: Just a thought, does maintaining constant values for exit status in this case makes sense? So instead of 124 if it said something like `exit_status == SIGTERM_EXIT_CODE` and then maintaining values of these exit codes in a common file maybe called as `constants.rb`? This way we can avoid hardcoding values in our code and attach meaning to it.

PR #690, opened 2021-06-04
Parse Message : Added new message for lack of sudo privilege and appl…

msys-sgarg:
Error message thrown  when user needs password everytime to perform a sudo operation on the target OS is not well handled currently 

## Description

Current Error message thrown to the user is like this : - 

```
ERROR: Train::UserError: Sudo failed: sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helpe
```
Handled in train code and returning a formatted message, that is easy to understand and helps user exactly what is the error like this : - 

```
Sudo does not have enough privileges for the operation, try passing password with -P or --ssh-password option
```
## Related Issue
https://github.com/chef/chef/issues/11359

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/extras/command_wrapper.rb
+++ b/lib/train/extras/command_wrapper.rb
@@ -81,6 +81,9 @@ module Train::Extras
       when /sudo: sorry, you must have a tty to run sudo/
         ["Sudo requires a TTY. Please see the README on how to configure "\
           "sudo to allow for non-interactive usage.", :sudo_no_tty]
+      when /sudo: a terminal is required to read the password; either use/
+        ["Sudo cannot prompt for password because there is no terminal. "\
+            "Please provide the sudo password directly", :sudo_missing_terminal]
       else
         [rawerr, nil]
       end
--- a/test/unit/extras/command_wrapper_test.rb
+++ b/test/unit/extras/command_wrapper_test.rb
@@ -121,28 +121,28 @@ describe "linux command" do
       backend.stubs(:run_command).returns(mock_connect_result("Sorry, try again", 1))
       lc = cls.new(backend, { sudo: true })
       err = _ { lc.verify! }.must_raise Train::UserError
-      _(err.message).must_match(/Sudo failed: Wrong sudo password./)
+      _(err.reason.to_s).must_match("bad_sudo_password")
     end
 
     it "error message for sudo password required" do
-      backend.stubs(:run_command).returns(mock_connect_result("sudo: no tty present and no askpass program specified", 1))
+      backend.stubs(:run_command).returns(mock_connect_result("sudo: no tty present and no program specified", 1))
       lc = cls.new(backend, { sudo: true })
       err = _ { lc.verify! }.must_raise Train::UserError
-      _(err.message).must_match(/Sudo requires a password, please configure it./)
+      _(err.message).must_match(/Sudo failed: sudo: no tty present and no program specified/)
     end
 
     it "error message for sudo: command not found" do
       backend.stubs(:run_command).returns(mock_connect_result("sudo: command not found", 1))
       lc = cls.new(backend, { sudo: true })
       err = _ { lc.verify! }.must_raise Train::UserError
-      _(err.message).must_match(/Can't find sudo command. Please either install and configure it on the target or deactivate sudo./)
+      _(err.reason.to_s).must_match("sudo_command_not_found")
     end
 
     it "error message for requires tty" do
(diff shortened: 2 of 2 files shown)


test/unit/extras/command_wrapper_test.rb
```diff
@@ -125,10 +125,10 @@ def mock_connect_result(stderr, exit_status)
     end
 
     it "error message for sudo password required" do
-      backend.stubs(:run_command).returns(mock_connect_result("sudo: no tty present and no askpass program specified", 1))
+      backend.stubs(:run_command).returns(mock_connect_result("sudo: no tty present and no program specified", 1))
       lc = cls.new(backend, { sudo: true })
       err = _ { lc.verify! }.must_raise Train::UserError
-      _(err.message).must_match(/Sudo requires a password, please configure it./)
+      _(err.message).must_match(/Sudo failed: sudo: no tty present and no program specified/)
```
marcparadise, line 131: it would be good to verify the `reason` field here and below too

lib/train/extras/command_wrapper.rb
```diff
@@ -81,6 +81,9 @@ def verify
       when /sudo: sorry, you must have a tty to run sudo/
         ["Sudo requires a TTY. Please see the README on how to configure "\
           "sudo to allow for non-interactive usage.", :sudo_no_tty]
+      when /sudo: a terminal is required to read the password; either use/
+        ["Sudo does not have enough privileges for the operation, "\
+         "try passing password with -P or --ssh-password option", :sudo_lack_of_permission]
```
marcparadise, line 86: I don't think this text  works coming out of train, with these specific options. This looks more like what we would want Chef Infra Client to be displaying.

msys-sgarg: ohh ok, so  this text will work - `Sudo does not have enough privileges for the operation. Need to pass password` ?

test/unit/extras/command_wrapper_test.rb
msys-sgarg: ohh ok,  will make that change

msys-sgarg: I have replaced message with reason in most test cases . but in few cases the `_err.reason` is coming to be blank string.I kept `_err.message` in those cases

test/unit/extras/command_wrapper_test.rb
```diff
@@ -1,11 +1,12 @@
 # author: Dominik Richter
 # author: Christoph Hartmann
 
-require "helper"
+require "helper.rb"
```
tas50:
```suggestion
require "helper"
```

msys-sgarg: fixed it right away

lib/train/extras/command_wrapper.rb
marcparadise: Looking closer, I think this is an inaccurate description of the problem. What is the permission that is missing? Isn't this failure about not being able to read from stdin because there's no terminal? 

In that case, maybe something like "Sudo cannot prompt for password because there is no terminal. Please provide the sudo password directly" ?

msys-sgarg: Ok sounds good and apt, i will change the message.

msys-sgarg: I have updated PR with correct error message, please check

lib/train/extras/command_wrapper.rb
```diff
@@ -81,6 +81,9 @@ def verify
       when /sudo: sorry, you must have a tty to run sudo/
         ["Sudo requires a TTY. Please see the README on how to configure "\
           "sudo to allow for non-interactive usage.", :sudo_no_tty]
+      when /sudo: a terminal is required to read the password; either use/
+        ["Sudo cannot prompt for password because there is no terminal. "\
+            "Please provide the sudo password directly", :sudo_lack_of_permission]
```
marcparadise, line 86: This looks good, but the symbol makes no sense any more - let's go with `:sudo_missing_terminal`?

marcparadise (changes requested):
Almost there! :) If you can change the symbol to match the error condition better, this is good to merge.

lib/train/extras/command_wrapper.rb
msys-sgarg: Did the change please check

marcparadise (approved):

PR #692, opened 2021-06-14
Add x25519 gem for SSH KEX support

### LINKED ISSUE #691: SSH Connection Fails When using non-default KEX
When SSHD is configured with a restricted set of key exchange algorithms, inspec/train fails to connect, with error:

```
inspec shell -t ssh://... -i ...
Transport error, can't connect to 'ssh' backend: SSH command failed (could not settle on kex algorithm)
```

The server is configured with KEX:
```
kexalgorithms curve25519-sha256,curve25519-sha256@libssh.org,diffie-hellman-group16-sha512,diffie-hellman-group18-sha512,diffie-hellman-group14-sha256
```

Seen on RHEL8, thought to be from an omnibus install, unknown inspec version.

clintoncwolfe:
Turns out that the `ed25519` gem only provides key algorithms, not key *exchange* algorithms, which are included in a separate sister gem. This bit us on inspec. When one is included, the other should be as well.

## Related Issue

Fixes #691

## Types of changes

## Checklist:

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :test do
   gem "pry-byebug"
   gem "m"
   gem "ed25519" # ed25519 ssh key support
+  gem "x25519" # ed25519 ssh key support for KEX
   gem "bcrypt_pbkdf" # ed25519 ssh key support
   # This is not a true gem installation
   # (Gem::Specification.find_by_path('train-gem-fixture') will return nil)


tas50:
Heads up before we add this to inspec. We had to back it out of chef/chef as it caused ruby to core dump in various platforms: https://github.com/RubyCrypto/x25519/issues/22

tarcieri:
FYI, the SIGILL errors should be fixed in `x25519` v1.0.9

PR #698, opened 2021-06-30
Fixed ssh connection issue if key_files array is empty

sanga1794:
Signed-off-by: sanga17 <sausekar@msystechnologies.com>

## Description
If key_files set as empty array then failing to ssh connection while bootstrapping the node

## Related Issue
https://github.com/chef/chef/issues/7053

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -177,7 +177,7 @@ module Train::Transports
         max_wait_until_ready: opts[:max_wait_until_ready],
         auth_methods: opts[:auth_methods],
         keys_only: opts[:keys_only],
-        keys: opts[:key_files].empty? ? nil : opts[:key_files],
+        keys: (!opts[:key_files].nil? && opts[:key_files].empty?) ? nil : opts[:key_files],
         password: opts[:password],
         forward_agent: opts[:forward_agent],
         proxy_command: opts[:proxy_command],
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -177,7 +177,7 @@ module Train::Transports
         max_wait_until_ready: opts[:max_wait_until_ready],
         auth_methods: opts[:auth_methods],
         keys_only: opts[:keys_only],
-        keys: opts[:key_files],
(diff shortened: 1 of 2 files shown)


tas50:
Is there a bug to link here and can we get a test added for this?

sanga1794:
> Is there a bug to link here and can we get a test added for this?

@tas I'm working on the issue https://github.com/chef/chef/issues/7053, while trying to bootstrap a node using the ssh config_file, as `key_files` from `chef` repo is `[]`(if we didn't mentioned in CLI), then not able to connect with the node using ssh.

sanga1794:
@tas50, @marcparadise Please review this PR Thanks!!

marcparadise:
How does having an empty array for key_files cause a failure?

sanga1794:
> How does having an empty array for key_files cause a failure?

@marcparadise below i attached the screenshot which consist of both failing and working `Net::SSH.start` method, please have look. Thanks!!
And we are deleting from list of options if value of any variable is nil for reference: https://github.com/inspec/train/blob/7121af0ea78734e4bf8081f14fb72a5f85378477/lib/train/transports/ssh_connection.rb#L213
Screenshot from 2021-07-28 15-52-28

sanga1794:
> > How does having an empty array for key_files cause a failure?
> 
> @marcparadise below i attached the screenshot which consist of both failing and working `Net::SSH.start` method, please have look. Thanks!!
> And we are deleting from list of options if value of any variable is nil for reference:
> 
> https://github.com/inspec/train/blob/7121af0ea78734e4bf8081f14fb72a5f85378477/lib/train/transports/ssh_connection.rb#L213
> 
> 
> Screenshot from 2021-07-28 15-52-28

@marcparadise, @tas50 Please have look into this. Thanks!!!

marcparadise:
If auth method were `public-key,keyboard-interactive` I'd then expect to receive a password prompt (if tty enabled). With just `public-key` if the key is not present, it can't succeed.   Based on that understanding, it seems like what you're seeing is correct. 

Is there something different that you'd expect, or am I misunderstanding something in that output?

PR #709, opened 2021-11-19
Adds ssh_config_file option.

### LINKED ISSUE #462: Support SSH Config file
## Detailed Description

Please support the SSH Configuration file.  This is essential functionality for those using Linux in data centers, whether there's a need to create ssh-config for ephemeral systems generated dynamically, or when uses non-default solution for a host name, whether different set of credentials, using jump servers, tunnels, etc.

Without support for SSH Configuration file, fundamental SSH features supported in train clients like Inspec are limited at best.

## Context

There are so many use cases where and why ssh-config file would be used.  It comes up often enough with ephemeral systems, such as test systems with Vagrant and Test Kitchen, Docker containers that support ssh, cloud services with AWS, Google Cloud, Azure, or for use cases where you need to use non-default credentials, such as build-deploy bot on CI/CD system that uses it for github.com host, or needed when tunnels or jump hosts are needed to get to secure systems.

Below is an example usage with train client Inspec, should we be able to have this feature: 

```bash
vagrant ssh-config > ssh-config # only need to do once
inspec exec test.rb -t ssh://systemA -F ssh-config
inspec exec test.rb -t ssh://systemB -F ssh-config
```

At this moment, because the train library doesn't support SSH Config, we have to do this below.  This these settings are dynamic, it is not possible to automate this easily. 

```bash
inspec exec test.rb \
  -t ssh://vagrant@localhost \
  -i $HOME/workspace/proj/.vagrant/machines/systemA/virtualbox/private_key \  
  -p 2222
inspec exec test.rb \
  -t ssh://vagrant@localhost \
  -i $HOME/workspace/proj/.vagrant/machines/systemB/virtualbox/private_key \  
  -p 2220
```

## Possible Implementation

*  Use the existing functionality already coded and done for you: `Net::SSH:Config` ([docs](https://net-ssh.github.io/net-ssh/Net/SSH/Config.html).

Vasu1105:

## Description

There is currently no option to set the ssh_config file path and the train was not reading all the values from the ssh config file so updated the code to add that option which can be set to true/false or to the path to the ssh config file. Also resolved the issue of all values from ssh config file is not getting set.

## Related Issue

Fixes #462  partly
## Types of changes

## Checklist:

### DIFF
--- a/lib/train/options.rb
+++ b/lib/train/options.rb
@@ -59,9 +59,6 @@ module Train
           default = hm[:default]
           if default.is_a? Proc
             res[field] = default.call(res)
-          elsif hm.key?(:coerce)
-            field_value = hm[:coerce].call(res)
-            res[field] = field_value.nil? ? default : field_value
           else
             res[field] = default
           end
--- a/lib/train/options.rb
+++ b/lib/train/options.rb
@@ -59,6 +59,9 @@ module Train
           default = hm[:default]
           if default.is_a? Proc
             res[field] = default.call(res)
+          elsif hm.key?(:coerce)
+            field_value = hm[:coerce].call(res)
+            res[field] = field_value.nil? ? default : field_value
           else
             res[field] = default
           end
--- a/lib/train/transports/ssh.rb
+++ b/lib/train/transports/ssh.rb
@@ -43,11 +43,11 @@ module Train::Transports
 
     # common target configuration
     option :host,      required: true
-    option :port,      default: 22, coerce: proc { |u| read_options_from_ssh_config(u, :port) }, required: true
-    option :user,      default: "root", coerce: proc { |u| read_options_from_ssh_config(u, :user) }, required: true
+    option :port,      default: 22, required: true
+    option :user,      default: "root", required: true
     option :key_files, default: nil
     option :password,  default: nil
-
+    option :ssh_config, default: true
     # additional ssh options
     option :keepalive, default: true
     option :keepalive_interval, default: 60
@@ -75,6 +75,7 @@ module Train::Transports
 
     # (see Base#connection)
     def connection(state = {}, &block)
+      apply_ssh_config(options[:host])
       opts = merge_options(options, state || {})
       validate_options(opts)
       conn_opts = connection_options(opts)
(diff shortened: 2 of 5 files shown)


lib/train/transports/ssh.rb
```diff
@@ -86,16 +87,30 @@ def connection(state = {}, &block)
...
-    def self.read_options_from_ssh_config(options, option_type)
-      config_options = Net::SSH.configuration_for(options[:host], true)
-      config_options[option_type]
+    def apply_ssh_config(host)
+      if options[:ssh_config] != false && !options[:ssh_config].nil?
+        files = options[:ssh_config] == true ? Net::SSH::Config.default_files : options[:ssh_config]
+        host_cfg = ssh_config_for_host(host, files)
+        host_cfg.each do |key, value|
+          if key == :keys && options[:key_files].nil? && !host_cfg[:keys].nil? && options[:password].nil?
+            options[:key_files] = host_cfg[key]
```
clintoncwolfe, line 96: I had to re-read this a few times and read the tests before I understood that this was simply remapping the private keys option. A comment to that effect would be helpful.

lib/train/transports/ssh.rb
```diff
@@ -86,16 +87,30 @@ def connection(state = {}, &block)
...
+          else
+            # Give precedence to config file when ssh_config options is set to true or to the path of the config file.
+            # This is required as there are default values set for some of the opitons and we unable to
+            # identify whether the values are set from the cli option or those are default so either we should give
+            # precedence to config file or otherwise we need to check each options default values and then set the value for that option.
```
clintoncwolfe, line 101: I understand why this is, but this is likely to cause some pain and confusion. We'll need to document this carefully in the consuming CLI applications.

lib/train/options.rb
```diff
@@ -59,9 +59,6 @@ def merge_options(base, opts)
           default = hm[:default]
           if default.is_a? Proc
             res[field] = default.call(res)
-          elsif hm.key?(:coerce)
-            field_value = hm[:coerce].call(res)
-            res[field] = field_value.nil? ? default : field_value
```
clintoncwolfe, line 64: With this change, the user can no longer override the port and user on the command line. I understand not wanting to provide that functionality for all options, but removing the capability when it is already been provided is a breaking change. We should try to preserve this ability.

lib/train/transports/ssh.rb
```diff
@@ -43,11 +43,11 @@ class SSH < Train.plugin(1) # rubocop:disable Metrics/ClassLength
...
     # common target configuration
     option :host,      required: true
-    option :port,      default: 22, coerce: proc { |u| read_options_from_ssh_config(u, :port) }, required: true
-    option :user,      default: "root", coerce: proc { |u| read_options_from_ssh_config(u, :user) }, required: true
+    option :port,      default: 22, required: true
+    option :user,      default: "root", required: true
     option :key_files, default: nil
     option :password,  default: nil
-
+    option :ssh_config, default: false
```
clintoncwolfe, line 50: I suggest renaming this option to be :ssh_config_file and adding a comment explaining that you may pass the value `true` to use the default location.

lib/train/transports/ssh.rb
```diff
@@ -75,6 +75,7 @@ class SSH < Train.plugin(1) # rubocop:disable Metrics/ClassLength
 
     # (see Base#connection)
     def connection(state = {}, &block)
+      apply_ssh_config(options[:host])
```
clintoncwolfe, line 78: I recommend renaming this method to be apply_ssh_config_file

clintoncwolfe (approved):
A few minor change requests, the biggest one being to preserve the ability to override the port and user options.

Additionally, I am concerned about the failing ruby 2.6 tests.

PR #726, opened 2022-03-15
CFINSPEC-127 Generate uuid for docker container.

Vasu1105:

## Description

It generates a UUID for the docker container. Default it will use the host id to generate the unique id.
## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/docker.rb
+++ b/lib/train/transports/docker.rb
@@ -84,7 +84,7 @@ class Train::Transports::Docker
     end
 
     def unique_identifier
-      uuid = @id # default uuid set to the docker host.
+      uuid = @container.nil? ? @id : @container.id # default uuid set to the docker host.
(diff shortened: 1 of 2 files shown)


lib/train/transports/docker.rb
```diff
@@ -83,6 +83,17 @@ def uri
       end
     end
 
+    def unique_identifier
+      uuid = @id # default uuid set to the docker host.
```
clintoncwolfe, line 87: This should default to `@container.id`, not `@id` - otherwise on windows we will have the same ID for all containers, that of the host.

Vasu1105: I am not sure why it will have the same id for all the containers on Windows. But yes I will add make the change to use @conainer.id if not @container.nil?

PR #730, opened 2022-05-18
Add more flexibility to authorization unique identifier expectations

rbclark:
The current code just assumes the unique identifier responds to client ID and issuer. That is currently the case however it is potentially a flawed assumption with new authentication methods being added going forward.

## Description
I am currently working to add support to the googleauth library for Workload Identity Federation. As part of this, I am adding a new type of authentication to the googleauth library itself. It is currently assumed that the authorization responds to client_id and issuer, and if either of these are not present then train exits with a stacktrace. This small changes makes train more flexible regarding the authorization object.

## Related Issue
https://github.com/googleapis/google-auth-library-ruby/issues/354

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/transports/gcp.rb
+++ b/lib/train/transports/gcp.rb
@@ -105,9 +105,10 @@ module Train::Transports
       def unique_identifier
         unique_id = "default"
         # use auth client_id for users (issuer is nil)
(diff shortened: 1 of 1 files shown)


rbclark:
Sorry to be a bother but would it be possible for someone to take a look at this and let me know if it needs any changes?

Vasu1105 (approved):
This looks good to me. @clintoncwolfe

PR #742, opened 2023-01-26
Fix upload to support individual files

thheinen:
## Description

This enables transport plugins, who do not implement or have native file transfer capabilities (e.g. `train-telnet` or `train-serial`) to transfer individual files instead of complete directories.

[SSH](https://github.com/inspec/train/blob/main/lib/train/transports/ssh_connection.rb#L119-L132), [WinRM](https://github.com/inspec/train-winrm/blob/main/lib/train-winrm/connection.rb#L81-L83), and [Local](https://github.com/inspec/train/blob/main/lib/train/transports/local.rb#L42-L48) bring their own implementations of this function and are not affected by the change

## Related Issue

Knife bootstrapping via compatible Train plugins, see chef/chef#13534

## Types of changes

## Checklist:

### DIFF
--- a/lib/train/plugins/base_connection.rb
+++ b/lib/train/plugins/base_connection.rb
@@ -162,24 +162,25 @@ class Train::Plugins::Transport
       @cache[:file][path] ||= file_via_connection(path, *args)
     end
 
-    # Uploads local files or directories to remote host.
+    # Uploads local files to remote host.
     #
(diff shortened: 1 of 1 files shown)


Vasu1105 (approved):
LGTM

tecracer-theinen:
Seems we need another review. The Knife part is moving now, so this will be a blocker if not merged.

PR #746, opened 2023-05-17
CHEF-2414 Deprecate azure connections from train

sathish-progress:
Warns the users that long unused Azure connections from train would be deprecated. 

## Description

As part of removing the long unused Azure SDKs from train, this change is the first part where we let know the users about deprecating this Azure Connection.
While target platform `azure` meaning `-t azure` would still continue to work.
Things that will be deprecated - 
- msi authentication
- any azure SDK based connections including graph rbac, vault, storage. These resources are long supported in [inspec-azure](https://github.com/inspec/inspec-azure/) cloud resource packs through non-SDK means using latest Azure APIs and they will continue to exist.

Screenshot 2023-07-25 at 5 11 54 PM

## Related Issue

second part: https://github.com/inspec/train/pull/745
https://chefio.atlassian.net/browse/CHEF-2414

## Types of changes

## Checklist:


ahasunos (approved):

clintoncwolfe (changes requested):
I'm going to ask that we do this without using ActiveSupport. We try to keep it out of train-core (and no, azure transport is not in train-core) but I think the only reason train has an active_support dependency was due to azure. I think active support dependencies need to be well-justified. Here we could get by with a warning.

clintoncwolfe (approved):

PR #747, opened 2023-06-23
Fix cannot find a UUID when connect using train with local transport inside docker container

Vasu1105:
## Description

When running Train using local transport insider docker container it enables to find the UUID.

Previously, that would raise an error `Cannot find a UUID` because all UUID finding method would fail.

This was breaking inspec CI tests under Ruby 3.1 - https://buildkite.com/chef-oss/inspec-inspec-inspec-6-verify/builds/216#0188e7c9-2312-4857-87a8-7d0d67f4673f

This solution is currently only for docker containers. 
## Types of changes

## Checklist:


clintoncwolfe (approved):
It's regrettable that this is Docker-specific to start, but we need to get a fix into the CI system ASAP. Other runtimes will continue to report no UUID.

PR #749, opened 2023-10-06
CHEF-6440(CHEF-7426): Adds audit log support

Vasu1105:
## Description

This PR adds the ability to configure the audit log option using train and if the audit log is enabled it generates the audit log files in the default location.

This includes the following changes

* By default audit log is disabled. Users can set the `enable_audit_log` option to `true` while creating train transport to enable it.
* User can use audit_log_location and audit_log_size options to set the location and size of the audit log file
* Currently we only do audit logging for commands executed using train connection object and logging file path that the user is trying to access or read. It also logs the file upload operations that are happening through train connection (Note currently audit log for file operations is not in a matured state so it will not work for transport which overrides the upload functionality or any other file related actions for example reading content of the file etc.).

Test using irb

```
bundle exec irb -I lib
require 'train'
t = Train.create("local")
c = t.connection
c.run_command("whoami")
```

audit_log_location is a required parameter so any library or plugin using train to pass the audit log location be they want to enable the audit log.

Pending: dev-docs
## Types of changes

## Checklist:

### DIFF
--- a/lib/train/options.rb
+++ b/lib/train/options.rb
@@ -41,7 +41,7 @@ module Train
         # should we keep it to $stdout.
         {
           enable_audit_log: { default: false },
-          audit_log_location: { required: true },
+          audit_log_location: { required: true, default: nil },
           audit_log_app_name: { default: "train" },
           audit_log_size: { default: 2000000 },
           audit_log_frequency: { default: "daily" },
--- a/lib/train/options.rb
+++ b/lib/train/options.rb
@@ -41,8 +41,8 @@ module Train
           enable_audit_log: { default: false },
           audit_log_location: { required: true, default: nil },
           audit_log_app_name: { default: "train" },
-          audit_log_size: { default: 2097152 },
-          audit_log_frequency: { default: "daily" },
+          audit_log_size: { default: nil },
+          audit_log_frequency: { default: 0 },
         }
       end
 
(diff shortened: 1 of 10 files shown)


lib/train/options.rb
```diff
@@ -31,9 +31,23 @@ def option(name, conf = nil, &block)
...
+        @default_options.merge!(default_audit_log_options)
         @default_options
       end
 
+      def default_audit_log_options
+        # TODO: What should be the default audit log location if any of the application using train does not set it?
+        # should we keep it to $stdout.
+        {
+          enable_audit_log: { default: false },
+          audit_log_location: { required: true, default: nil },
```
clintoncwolfe, line 44: OK, yeah, that required: true broke all the tests. We can compromise.

How about set it to a path like ~/.chef/logs/train-audit.log - Use a peculiar train-specific default, and any reasonable application would have to override it. Does that work?

Vasu1105: I think that should work.

Vasu1105: @clintoncwolfe I think we can keep the required. We are able to separate our validation implementation for the audit log from the existing validate_option method, and we separated the audit log options from the current default_options so that those can be handled separately. This I think will make this change less breaking.

Nik08 (approved):

PR #757, opened 2023-11-13
Upgrade Google REST API Client to latest

balasubramanian-s:
Upgrade Google API Client

## Description

Replaces the monolith client with the service specific client gems
## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New content (non-breaking change)
- [x] Breaking change (a content change which would break existing functionality or processes)

## Checklist:

- [ ] I have read the **CONTRIBUTING** document.

PR #762, opened 2023-12-11
CHEF-1895- Update googleauth to support external account type authorisation in GCP

balasubramanian-s:
Update googleauth dependency to support external account type
## Description

This issue occurs because the train repository was using a old version of googleauth
Fix would be updating the googleauth gem in train repository but it has a dependency with google-api-client.
Since google-api-client monolith gem is deprecated, migrated the google-api-client in order to support external account type of authorisation.
## Types of changes

## Checklist:


balasubramanian-s:
Fixes:
https://github.com/inspec/train/pull/757 - Upgrading google auth library was implemented.

PR #787, opened 2025-04-08
CHEF-19301-Ruby 3.4.x Upgrade

balasubramanian-s:
- Ruby 3.4.x Upgrade
- Dropping support for Ruby 3.0
## Description

- [x] Fixes for warnings
- [x] Update pipelines for 3.4.x
- [x] Update ruby min version to 3.1

 - [ ] Please merge https://github.com/chef/chef-licensing/pull/197 PR before merging this PR. 
 - [ ] Merge after dropping support of Ruby 3.0 from InSpec

## Types of changes

## Checklist:


train.gemspec
```diff
@@ -52,4 +52,6 @@ Gem::Specification.new do |spec|
   spec.add_dependency "google-apis-cloudresourcemanager_v1", "~> 0.35.0"
   spec.add_dependency "google-apis-storage_v1", "~> 0.30.0"
   spec.add_dependency "google-apis-iam_v1", "~> 0.50.0"
+  # Gem dependency needed with Ruby 3.4 upgrade
+  spec.add_dependency "ostruct", "~> 0.1.0"
```
ahasunos, line 56: Could we add this dependency in train-core.gemspec? Maybe we can wrap it in a Ruby version check like this, since it’s only needed for Ruby 3.4+:

```ruby
if Gem::Version.new(RUBY_VERSION) >= Gem::Version.new("3.4.0")
  spec.add_dependency "ostruct", "~> 0.1.0"
end

```

ahasunos (approved):

PR #795, opened 2025-06-10
CHEF-19255 Fix named pipes to be more secure

sa-progress:
## Description

This PR implements security best practices for Windows named pipe usage in the local transport.

* Randomizes pipe names
* Sets strict ACLs on named pipes
* Verifies pipe ownership before connecting
* Ensures fully qualified paths are used for PowerShell execution

## Types of changes

* [x] Bug fix (non-breaking change which fixes an issue)
* [ ] New content (non-breaking change)
* [ ] Breaking change (a content change which would break existing functionality or processes)

## Checklist

* [x] I have read the **CONTRIBUTING** document.


lib/train/transports/local.rb
```diff
@@ -230,6 +231,12 @@ def acquire_pipe
 
           pipe = nil
 
+          # 4. Verify ownership before connecting
+          owner, current_user, is_owner = pipe_owned_by_current_user?(pipe_name)
+          unless is_owner
+            raise PipeError, "Unauthorized user '#{current_user}' tried to connect to pipe '#{pipe_name}'. Pipe is owned by '#{owner}'."
+          end
```
sathish-progress, line 238: but we just created with the right privileges .. why do we have to check this?

sa-progress: As recommended in the shared remediation, even if we create the pipe with the correct privileges and ACLs, verifying ownership before connecting is a critical defense-in-depth measure. This helps prevent race condition and pre-creation attacks, ensuring that our process only connects to pipes it actually owns.

lib/train/transports/local.rb
```diff
@@ -241,25 +248,25 @@ def acquire_pipe
...
+        # 5. Set strict ACLs on named pipes (PowerShell)
         def start_pipe_server(pipe_name)
           require "win32/process"
-
           script = <<-EOF
             $ErrorActionPreference = 'Stop'
-
-            $pipeServer = New-Object System.IO.Pipes.NamedPipeServerStream('#{pipe_name}')
+            $pipeSecurity = New-Object System.IO.Pipes.PipeSecurity
+            $rule = New-Object System.IO.Pipes.PipeAccessRule("#{ENV["USERNAME"]}", "FullControl", "Allow")
```
Nik08, line 257: Will the `USERNAME` be set by the user themselves, or is there something I am missing here? 

In case we are asking for USERNAME from the user:
- Not sure if we should ask user for this.
- Also for this logic to work, ENV value of username should match with `owner` and `current_user` values right?
And `current_user` and `owner` might return values in the format `"DOMAIN\\Username"`. Just making sure that we communicate this to the user to set USERNAME is in this format.

lib/train/transports/local.rb
```diff
@@ -221,7 +221,8 @@ def close
 
         def acquire_pipe
           require "win32/process"
-          pipe_name = "inspec_#{SecureRandom.hex}"
+          # 3. Randomize pipe name
+          pipe_name = SecureRandom.hex
```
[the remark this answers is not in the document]
Nik08: @sa-progress Since we are already checking the ownership of the pipe and also starting a pipe server with proper security, do we still need to make random names for pipes? Keeping `inspec_` does provide us with some control over debugging. Just checking if this was the conclusion we had reached and still need.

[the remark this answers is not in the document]
sa-progress: We can keep the pipe name same as inspec_*. if it helps in debugging.

